PR #34, opened 2015-02-11
90%: Sidestep cursor problems

rsinger:
This fixes a problem with Mongos vs. 2.6.+ where using 
```php 
while($cursor->hasNext()) {
    $result = $cursor->getNext();
```
causes ``` $result ``` to be empty.  It's not clear if this is a bug in php-mongo or mongo, but we've seen it consistently in 2.6+.  This should be backwards compatible with any versions of Mongo 2.0+


malcyL:
:+1:

lordtatty:
:+1:

PR #35, opened 2015-02-16
90%: Scripts redesign for getopts/stats

rsinger:


rsinger:
90% (updated)

rsinger:
It's temporary, until we can find all of the legacy places that are using the old scripts.

rsinger:
But, yes, the intention is to remove createViews once we've hunted them all down.

rsinger:
Oops, it's supposed to be merged into the rest-interface branch... closing.

PR #36, opened 2015-02-16
90%: Scripts redesign

rsinger:
This PR is the start of deprecating the old tripod scripts for generating views, tables, etc. so we can use getopts (rather than argument order) and include stats monitoring, etc.

Currently, they run alongside the old scripts until we can identify and remove all references to the old ones.

### DIFF
--- a/scripts/mongo/createTables.php
+++ b/scripts/mongo/createTables.php
@@ -27,7 +27,7 @@ function generateTables($id, $tableId,$storeName)
         }
         else
         {
-            print " for all views....\n";
+            print " for all tables....\n";
             $tTables->generateTableRows($tableId);
         }
     }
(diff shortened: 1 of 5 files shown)


rsinger:
Handwavy example of what the file the ``` --stat-loader ``` argument should look like:
```php
<?php

require_once './test/performance/rest-interface/src/Stat.class.php'; // Example iTripodStat based class
$storename = isset($options['s']) ? $options['s'] : $options['storename']; // or MongoTripodConfig::getInstance()->getQueueConfig()['database'] for queueProcess
StatConfig::setConfig(array('host'=>'127.0.0.1','port'=>8125));
return new Stat($storename);
```

malcyL:
:+1:

PR #37, opened 2015-02-17
90%: Create multiple repsets for testing

rsinger:
This adds utility scripts to create and start multiple replica sets on a local machine for testing.


malcyL:
:+1:

PR #38, opened 2015-02-17
10% Performance Test Tripod Queue

malcyL:
A 10% pull request - TestTripodQueue is working - but I'm not sure it should even be here. Perhaps it should be in https://github.com/talis/tripod-php-perf-test-driver

### DIFF
--- a/scripts/mongo/TestTripodQueue.class.php
+++ b/scripts/mongo/TestTripodQueue.class.php
@@ -1,0 +1,33 @@
+<?php
+
+require_once TRIPOD_DIR.'/mongo/queue/MongoTripodQueue.class';
+
+class TestTripodQueue extends MongoTripodQueue {
+
+    protected $template = null;
+
+    public function __construct($stat=null, $template) {
+        parent::__construct($stat);
+        $this->template = $template;
+    }
+
+    protected function getMongoTripod($data) {
+        setConfig($data);
+        parent::getMongoTripod($data);
+    }
+
+    public function setConfig($opts)
+    {
+        $c = $this->template;
+
+        $c = str_replace('#SHORT_CODE#',$opts["tenantShortCode"],$c);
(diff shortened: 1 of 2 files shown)


rsinger:
This isn't quite the right place.  I think it's fine to have the TestTripodQueue class, but it probably needs to be confined into test/performance somewhere.  Also processQueue2 shouldn't need to require it or instantiate it.  Instead, either put something in test/performance/rest-interface/utils or test/performance/rest-interface/scripts (which would need to be created).  This seems specific to the test suite, but not tripod itself.

scripts/mongo/TestTripodQueue.class.php
```diff
@@ -0,0 +1,47 @@
...
+
+class TestTripodQueue extends MongoTripodQueue {
+
+    protected $mapDatabaseToBaseUri = array(
+        "kent" => "http://resourcelists.kent.ac.uk",
+        "mmu"  => "http://lists.lib.mmu.ac.uk",
+        "ntu"  => "http://resourcelists.ntu.ac.uk",
+        "sta"  => "http://resourcelists.st-andrews.ac.uk",
+        "worc" => "http://readinglists.worcs.ac.uk"
+    );
```
robotrobot: Should't this be passed to the constructor? Seem really specific to our test runs.

rsinger: Yeah, this is being moved to our other repo.

malcyL: Absolutely. I created a 10% pull request to get some visibility of this, but see my comment from above:

```
A 10% pull request - TestTripodQueue is working - but I'm not sure it should even be here.
Perhaps it should be in https://github.com/talis/tripod-php-perf-test-driver
```

malcyL:
Closing as this is moving to the driver project.

PR #40, opened 2015-02-24
90%: Add computed fields

rsinger:
This is a spike to throw up a proof of concept to allow generation of dynamic table values based on the values elsewhere in the generated table row document.

It also introduces the notion of ‘temp’ fields that aren’t stored in the collection: these fieldNames have a ``` "temporary": true ``` property.

### Example of a table spec with this ###
```javascript
{
    "_id": "t_wibble",
    "type": "owl:Thing",
    "from": "CBD_foobar",
    "fields" : [
        {
            {
                "fieldName": "foobarLink",
                "value": "link"
            },   
            {
                "fieldName": "mainType",
                "predicates": ["rdf:type"],
                "temporary": true
            },
            {
                "fieldName": "lastModMain",
                "predicates": ["dct:modified"],
                "temporary": true
            },
            {
                "fieldName": "mainTitle",
                "predicates": ["dct:title"],
                "temporary": true
            }
        }
    ],
    "joins": {
        "dct:hasPart" : {
            "fields" : [
                {
                    "fieldName": "partTitle",
                    "predicates": ["dct:title"],
                    "temporary": true
                },
                {
                    "fieldName": "partType",
                    "predicates": ["rdf:type"],
                    "temporary": true
                },
                {
                    "fieldName": "lastModPart",
                    "predicates": ["dct:modified"],
                    "temporary": true
                }
            ]
        },
        "dct:isPartOf" : {
            "fields" : [
                {
                    "fieldName": "partOfTitle",
                    "predicates": ["dct:title"],
                    "temporary": true
                },
                {
                    "fieldName": "partOfType",
                    "predicates": ["rdf:type"],
                    "temporary": true
                },
                {
                    "fieldName": "lastModPartOf",
                    "predicates": ["dct:modified"],
                    "temporary": true
                }
            ]
        },        
    },
    "counts": [
        {
            "fieldName": "referenceCount",
            "property": "dct:references"
        },
        {
            "fieldName": "referencedCount",
            "property": "dct:isReferencedBy"
        }
    ],
    "computed_fields": [
        {
            "fieldName": "title",
            "value": {
                "conditional": {
                    "if":[["bibo:Article", "bibo:Chapter"], "contains", "$mainType"],
                    "then":"$mainTitle",
                    "else": {
                        "conditional":{
                            "if":[["bibo:Article", "bibo:Chapter"], "contains", "$partType"],
                            "then":"$partTitle",
                            "else":{"conditional":{
                                "if":[["bibo:Article", "bibo:Chapter"], "contains", "$partOfType"],
                                "then":"$partOfTitle",
                                "else":"Unknown!!"
                            }}
                        }
                    }
                }
            }
        },
        {
            "fieldName": "type",
            "value": {
                "replace": {
                    "search":"bibo:",
                    "replace":"",
                    "subject":"$mainType"
                }
            }
        },
        {
            "fieldName": "referenceDiff",
            "value": {
                "conditional": {
                    "if":["$referenceCount",">","$referencedCount"],
                    "then":{
                        "arithmetic":["$referenceCount","-","$referencedCount"]
                    },
                    "else":{
                        "arithmetic":["$referencedCount","-","$referenceCount"]
                    }
                }
            }
        }
    ]            
}
```
### Example of document outputted above ###
```javascript
{
    "foobarLink": "http://example.com/1234",
    "referenceCount": 32,
    "referencedCount": 14,
    "title": "I was the isPartOf title",
    "type": "Book",
    "referenceDiff": 18
}
```


src/mongo/delegates/MongoTripodTables.class.php
```diff
@@ -378,6 +407,310 @@ public function generateTableRows($tableType,$resource=null,$context=null)
...
+                break;
+            case "/":
+                $value = $left / $right;
+                break;
+            case "%":
+                $value = $left % $right;
+                break;
+            default:
+                $value = null;
+        }
```
malcyL: Arithmetic operators contains: "+", "-", "*", "/", "%", "**"

Did we miss "**" from this switch?

malcyL:
Reading through this, I was thinking there is a lot of parsing code - is there some library in PHP that might do this for us? A PHP yacc/bison? I guess we are already parsing the tablespec like this, and this is just extending what we have.

So - other than my question above on "**" - :+1:

src/mongo/delegates/MongoTripodTables.class.php
rsinger: No - it should have been removed from the valid operators: exponents were introduced in PHP 5.6, so I don't think we should support them here.

rsinger:
I was also concerned about the size of the validation code (which I think is what you mean about parsing), especially given that it needs to be done *on every request*.  I sort of wonder about keeping them, but making the actual bit that runs on MongoTripod::loadConfig() much smaller and put a more comprehensive one as a script.

rsinger:
The only downside to this is that if your tripod config is generated on the fly (like it is in TARL).

lordtatty:
My first thoughts on this is that we seem to have an element of "code is data, and data is code" hovering around here, and this is a problem which has already been solved by languages such as LISP.  I have found a couple of sites where people are discussing embedding conditionals within JSON, like we are trying to do, and lisp-like s-expressions seem to be approaches which are favoured, as they directly translate to this problem.  http://alandipert.tumblr.com/post/7193534410/discover-lisp-in-your-web-browser-with-javascript - part way down this page the author proposes a JSONscript language (I googled - it doesn't actually exist yet), but I wonder if it might provide some interesting ideas for us here.  Perhaps we could write a simple parser for JSONscript and open source it.  There is a stack overflow post here where someone is proposing a similar design:  http://stackoverflow.com/questions/20737045/representing-logic-as-data-in-json

Having said that, there is nothing inherently wrong with this approach if we are too far down the rabbit hole to change it now - I was just thinking in terms of standing on the shoulders of giants, and such.  It would also help with the above thoughts on validation by separating out that code from tripod itself.

rsinger:
@lordtatty Are you proposing that the "value" value for the fieldName "type" should look something like:
```javascript
["_replace_",["bibo","","$mainType"]]
```
and the "value" value for fieldName "referenceDiff" should look like:
```javascript
[
    "_conditional_",
    [
        "if",
        [
            ">",
            ["$referenceCount", "$referencedCount"]
        ],
        "then",
        [
            "_arithmetic_",
            [
                "-",
                ["$referenceCount","$referencedCount"]
            ]
        ],
        "else",
        [
            "_arithmetic_",
            [
                "-",
                ["$referencedCount","$referenceCount"]
            ]
        ]
    ]
]
```

lordtatty:
It's been a little while since I touched something like lisp myself, but for the latter I was thinking something along the lines of:

```
[
    "_conditional_", 
        ["-",
        ["MAX", "$referenceCount", "$referencedCount"],        
        ["MIN", "$referenceCount", "$referencedCount"],               
    ]
]
```
Maybe could make it easier a function like MAXMIN so we don't need to do two separate MAX and MIN calls, but in terms of raw functionality, something like that.

malcyL:
:+1:

rsinger:
So, I think these suggestions are good and fairly trivial to implement (incrementally).

We do have to decide how we want to reconcile consolidating the "predicate functions" and "computed field functions", though, since they operate on somewhat different principles which complicates defining a common interface for them all.

The main issue is that ``` _link_ ``` and the predicate functions work with the mongograph documents, while the computed fields (currently) operate on the expectation that table data will be provided.

There are a couple of ways to approach this:
1. The ``` ComputedFieldFunction ``` interface defines different methods for graph and tabular data: this would provide backwards compatibility, but introduces some complications.  config specifications (and validation thereof) could get complicated if there's a mix of nested graph and table computations.  Also, it raises the burden of introducing new functions, since they'd need to account for either graph or table data.  However, going this way would mean we could use this functionality in views, perhaps.
2. We could deprecate the inline functions and force that functionality to be moved to computed fields.  There's value in this approach: ``` _link_ ``` only works on the current ID, for example.  It also keeps the interface simple.  The downside is we would need to figure out how this would work in a view.

View compatibility may be a red herring: tables and search docs are flat arrays - implementing computed values for them is pretty simple.  Views complicate things quite dramatically: not only is their shape different (with uris and literals, for example -- and even more complicated if we ever introduce typed literals), but also multidimensional (in that you'll need a way to define graph traversal, possibly).

I'm not sure that, at this pass, we worry about graph data.  With that in mind, I opt for option 2, but I'm not opposed to 1.

rsinger:
I think all of the comments here have been addressed:
* Docs updated to include computed field functions and temporary fields
* ``` _function_ ``` has been removed in favor of just 'conditional', 'replace', etc.  '_link_' has been deprecated in favor of 'link' and now logs a warning.  Also log warning was added
* #44 
* The naming convention for temporary fields has been replaced with adding a ``` 'temporary': true ``` property on the fieldspec

docs/primers/tables.md
```diff
@@ -9,6 +9,264 @@ Tablespecs
...
+                    } // examples: join [combine multiple values into a string], lowercase, date [turns a string date into a MongoDate]
+                ]
+            }
+        }
+     },
+     "computed_fields": [ // this can only be at the top level of the tablespec, as it's applied after all other values are retrieved
+        {
+            "fieldName": "type",
+            "value": {
+                "_replace_" : { // _replace_ is a function name
```
rgubby: Should the underscores be removed?

src/mongo/MongoTripodConfig.class.php
```diff
@@ -296,6 +318,14 @@ protected function loadConfig(Array $config)
                 {
                     throw new MongoTripodConfigException("View spec does not contain " . _ID_KEY);
                 }
+                if(!isset($spec['from']) || !in_array($spec['from'], $this->getPods($storeName)))
```
rgubby: You've got a ```validateTableSpec``` function to validate table specs, but validation of view specs (and search specs) are done inline. Worth moving to a function to keep it consistent?

src/mongo/MongoTripodConfig.class.php
```diff
@@ -316,22 +346,70 @@ protected function loadConfig(Array $config)
...
-                {
-                    throw new MongoTripodConfigException("Table spec does not contain " . _ID_KEY);
-                }
+                $this->validateTableSpec($spec);
 
-                // Get all "fields" in the spec
-                $fieldsInTableSpec = $this->findFieldsInTableSpec('fields', $spec);
-                // Loop through fields and validate
-                foreach($fieldsInTableSpec as $field)
+                if(isset($spec['to_data_source']))
```
rgubby: Should this be in the ```validateTableSpec``` function?

src/mongo/MongoTripodConfig.class.php
rsinger: Originally, I had done this, but, actually, the amount of validation done for search and view specs is the same as we do inline for table specs.

Really, it's just that we do *a lot* more validation for table specs.

src/mongo/MongoTripodConfig.class.php
rsinger: No, because the validation of the table spec stuff is independent of the configuration as a whole.

src/mongo/base/MongoTripodBase.class.php
```diff
@@ -206,6 +206,25 @@ public function errorLog($message, $params=null)
         }
     }
 
+    public function warningLog($message, $params=null)
```
rgubby: missing docblock

src/mongo/base/MongoTripodBase.class.php
```diff
@@ -206,6 +206,25 @@ public function errorLog($message, $params=null)
...
+    {
+        if (self::getLogger()!=null)
+        {
+            self::getLogger()->getInstance()->warn("[TRIPOD_WARN] $message",$params);
+        }
+        else
+        {
+            echo "$message\n";
```
rgubby: ```echo``` or ```this->debugLog``` ?

docs/primers/tables.md
```diff
@@ -9,6 +9,264 @@ Tablespecs
...
+            "predicates": ["dct:title"] // Every field must have a 'predicates' or 'value' property
+        },
+        {
+            "fieldName": "resourceLink",
+            "value": "link" // This turns the subject's ID into a fully qualified uri
+        },
+        {
+            "fieldName": "typeQname",
+            "predicates": ["rdf:type"],
+            "temporary": true  // fieldNames with the temporary property set to true will not be stored in the document
```
robotrobot: Might just be worth noting here *why* you want a temporary field

rsinger: Keep reading.

docs/primers/tables.md
```diff
@@ -9,6 +9,264 @@ Tablespecs
...
+```
+
+If the array's count is three, the 2nd value must be one of:
+">","<",">=", "<=", "==", "!=", "contains", "not contains", "~=", "!~"
+there is no strict type comparison.
+
+"then" and "else" values may be of any type, including another function.  To replicate 'else if', you can use another
+ conditional function as the value of 'else'.
+
+ Arithmetic functions
```
rgubby: Do you want something in here to tell you what arithmetic functions are available?

src/mongo/delegates/MongoTripodTables.class.php
```diff
@@ -378,6 +414,332 @@ public function generateTableRows($tableType,$resource=null,$context=null)
...
+                        $value = (int)$value;
+                    }
+                    else
+                    {
+                        $value = (float)$value;
+                    }
+                }
+                break;
+        }
+        return $value;
```
rgubby: If type is null what happens? If you're allowing type to be null, do you need something to try and work out what type the value is?

rsinger:
```
php> (string)null
''
php> (int)null
int(0)
php> 
```

docs/primers/tables.md
```diff
@@ -9,6 +9,267 @@ Tablespecs
...
+If, for this example, ``` $x ``` is "bibo:Article"
+
+```javascript
+{
+    "fieldName": "foo",
+    "value": {
+        "replace": {
+            "search": "bibo:",
+            "replace": "",
+            "subject": "$x"
```
rgubby: I noticed in the PR that uses this functionality that you have something like:
```"subject": "$!concatenatedType"``` - what does that do?

Comment (rgubby) [inline docs/primers/tables.md]: I think that "!concatenatedType" is the name of the field - just confused me a bit as it looked like it could be an operator..

Comment (rsinger) [inline docs/primers/tables.md]: That was based on an earlier version of this branch.

Comment (rsinger) [inline docs/primers/tables.md]: READ THE COMMENTS, @rgubby!

File: src/mongo/delegates/MongoTripodTables.class.php [hunk 11]
```diff
@@ -26,6 +26,35 @@ class MongoTripodTables extends MongoTripodBase implements SplObserver
...
+     */
+    public static $computedFieldFunctions = array('conditional', "replace", "arithmetic");
+
+    /**
+     * Computed conditional config - list of allowed conditional operators
+     *
+     * @var array
+     * @static
+     */
+    public static $conditionalOperators = array(">","<",">=", "<=", "==", "!=", "contains", "not contains");
```
rgubby: Should this also contain ```~=``` and ```!~``` ?

test/performance/rest-interface/config/tripod-config.json
```diff
@@ -646,6 +646,50 @@
...
                         }
                     ]
+                },
+                {
+                    "_id": "t_conditional_creators",
+                    "type": ["bibo:Book", "bibo:Document"],
+                    "from": "CBD_testing",
+                    "counts": [
+                        {
+                            "fieldName": "!creatorCount",
```
rgubby: Is there a reason why the field names have a ```!``` in front of them? It looks a bit confusing to me - "not the creatorCount"?

test/performance/rest-interface/config/tripod-config.json
```diff
@@ -646,6 +646,50 @@
...
+                        {
+                            "fieldName": "!contributorCount",
+                            "property": "dct:contributor"
+                        }
+                    ],
+                    "computed_fields": [
+                        {
+                            "fieldName": "!creatorCount",
+                            "value": {
+                                "_conditional_": {
```
rgubby: should this have underscores?

test/performance/rest-interface/config/tripod-config.json
```diff
@@ -646,6 +646,50 @@
...
+                                    "if":["$!creatorCount"],
+                                    "then":"$!creatorCount",
+                                    "else": 0
+                                }
+                            }
+                        },
+                        {
+                            "fieldName": "!contributorCount",
+                            "value": {
+                                "_conditional_": {
```
rgubby: should this have underscores?

test/performance/rest-interface/config/tripod-config.json
```diff
@@ -646,6 +646,50 @@
...
+                                    "if":["$!contributorCount"],
+                                    "then":"$!contributorCount",
+                                    "else": 0
+                                }
+                            }
+                        },
+                        {
+                            "fieldName": "totalContributorCount",
+                            "value": {
+                                "_arithmetic_":["$!creatorCount","+","$!contributorCount"]
```
rgubby: should this have underscores?

rgubby:
looks good to me! :+1:

PR #43, opened 2015-03-20
90%: Add read retries to config to add more load options

rsinger:
This adds an option to repeat reads multiple times to help load testing.


test/performance/rest-interface/config/config.json
```diff
@@ -10,5 +10,6 @@
             "generate_search_index_docs" : true
         },
         "readPreference":"primaryPreferred"
-    }
+    },
+    "read-retries": 0
```
robotrobot: Not really a retry. More a multiplier.

PR #48, opened 2015-04-24
90% - Use correct config array when getting the transaction log db

lordtatty:
Tripod is currently using the queue db to store transaction logs.  Fixed this, and added a test to ensure queues and transaction logs stick to their own dbs.

### DIFF
--- a/src/mongo/MongoTripodConfig.class.php
+++ b/src/mongo/MongoTripodConfig.class.php
@@ -2047,7 +2047,7 @@ class MongoTripodConfig
             $connectionOptions['replicaSet'] = $dataSource['replicaSet'];
         }
         $client = new MongoClient($dataSource['connection'], $connectionOptions);
-        $db = $client->selectDB($this->queueConfig['database']);
+        $db = $client->selectDB($this->tConfig['database']);
         $db->setReadPreference($readPreference);
         return $db;
     }
--- a/test/unit/mongo/MongoTripodConfigTest.php
+++ b/test/unit/mongo/MongoTripodConfigTest.php
@@ -1399,12 +1399,16 @@ class MongoTripodConfigTest extends MongoTripodTestBase
         $queueDB->drop();
 
         // Make sure the dbs do not exist
-        $transactionMongo = new MongoClient($newConfig['data_sources'][$newConfig['transaction_log']['data_source']]['connection']);
+        $transactionConnInfo = $newConfig['data_sources'][$newConfig['transaction_log']['data_source']];
+        $options = isset($transactionConnInfo['replicaSet']) && !empty($transactionConnInfo['replicaSet']) ? array('replicaSet' => $transactionConnInfo['replicaSet']): array();
+        $transactionMongo = new MongoClient($transactionConnInfo['connection'], $options);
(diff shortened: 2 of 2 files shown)


src/mongo/MongoTripodConfig.class.php
```diff
@@ -2047,7 +2047,7 @@ public function getTransactionLogDatabase($readPreference = MongoClient::RP_PRIM
             $connectionOptions['replicaSet'] = $dataSource['replicaSet'];
         }
         $client = new MongoClient($dataSource['connection'], $connectionOptions);
-        $db = $client->selectDB($this->queueConfig['database']);
+        $db = $client->selectDB($this->tConfig['database']);
```
kiyanwang: thats definitely the culprit

test/unit/mongo/MongoTripodConfigTest.php
```diff
@@ -1380,6 +1380,78 @@ public function testDataLoadedInConfiguredDataSource()
         }
     }
 
+    public function testTransactionLogAndQueuesAreWrittenToCorrectDBAndCollection()
```
kiyanwang: that test looks reasonable.
- make sure theres nothing in the db
- do a write
- make sure its written into the correct dbs

test/unit/mongo/MongoTripodConfigTest.php
```diff
@@ -1380,6 +1380,78 @@ public function testDataLoadedInConfiguredDataSource()
...
+        $graph->add_resource_triple($subject, RDF_TYPE, $labeller->qname_to_uri('foaf:Person'));
+        $graph->add_literal_triple($subject, FOAF_NAME, "Anne Example");
+        $this->tripod->saveChanges(new ExtendedGraph(), $graph);
+
+        $newGraph = $this->tripod->describeResource($subject);
+        $newGraph->add_literal_triple($subject, $labeller->qname_to_uri('foaf:email'), 'anne@example.com');
+        $this->tripod->saveChanges($graph, $newGraph);
+
+        // Make sure the dbs do now exist
+        $mongo = new MongoClient(MongoTripodConfig::getInstance()->getConnStr('tripod_php_testing'));
```
rsinger: This won't work - the transactions and queue might not use the same connection string

rsinger: I guess what I'm saying is that these need to sanity check from the config, not from calling Mongo

lordtatty: How do I get the correct connection string in this instance?

rsinger: If you want to be sure that it's written to the database that was defined in the config, you're going to have to get that from the config document.  For the tlog, you'd have to do something like:
```php 
$tlogDataSource = $config['data_sources'][$config['transaction_log']['data_source']]
```
then, if $tlogDataSource['type'] == 'mongo', you can use the 'connection' and 'replicaSet' properties to init MongoClient with

Queue would be basically the same:
```php
$queueDataSource = $config['data_sources'][$config['queue']['data_source']]
```

rsinger:
:+1:

PR #51, opened 2015-04-28
30% - Resqueize queue

robotrobot:
Replaces the custom mongo queue with a resque one.

Because config is now passed to a job completely negates the need for a custom queue runner in our app.

Todo:
* ~~Passing tests~~
* We really need to get this in the lab as the amount of jobs results in a lot of db connections potentially. Want to assess the impact at scale of this.


scripts/mongo/worker.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+require "../../src/tripod.inc.php";
+$logger = new \Psr\Log\NullLogger(); // silence resque // todo: implement PSR logger across Tripod.
```
robotrobot: It's becoming plan that we'll need a PSR logger, especially if we want to log the results of resque jobs properly.

src/mongo/ModifiedSubject.class.php
```diff
@@ -6,13 +6,13 @@
  */
 class ModifiedSubject implements SplSubject
```
robotrobot: Doubting the usefulness of an SplSubject now observer->subject is a 1->1

src/mongo/MongoTripodConstants.php
```diff
@@ -64,7 +64,8 @@
 define('MONGO_LOCK','MONGO_LOCK');
 define('MONGO_ADD_TO_QUEUE','MONGO_ADD_TO_QUEUE');
 define('MONGO_QUEUE_FAIL','MONGO_QUEUE_FAIL');
-define('MONGO_QUEUE_SUCCESS','MONGO_QUEUE_SUCCESS');
+define('MONGO_QUEUE_DISCOVER_SUCCESS','MONGO_QUEUE_DISCOVER_SUCCESS');
+define('MONGO_QUEUE_APPLY_OPERATION_SUCCESS','MONGO_QUEUE_APPLY_OPERATION_SUCCESS');
```
robotrobot: queue is now two distinct elements so retiring MONGO_QUEUE_SUCCESS

src/mongo/base/JobBase.class.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function debugLog($message, $params = null)
+    {
+        parent::debugLog("[PID ".getmypid()."] ".$message, $params);
+    }
+
+    public function errorLog($message, $params = null)
+    {
+        parent::errorLog("[PID ".getmypid()."] ".$message, $params);
+    }
```
robotrobot: More evidence for needing `PsrLogger`

src/tripod.inc.php
```diff
@@ -1,20 +1,37 @@
 <?php
-if(!defined('TRIPOD_DIR')) define('TRIPOD_DIR', dirname(__FILE__) . '/' );
-
-if (version_compare(phpversion(), '5.3.0', '<')) {
```
robotrobot: bye bye < 5.3.0. Die you MF.

test/unit/mongo/MongoTripodQueueOperationsTest.php
```diff
@@ -11,8 +10,9 @@
  * Tripod Queue, and for each of those items added to the queue the correct operations are listed; furthermore in some cases that when operations are performed
  * the results are as we would expect. For that reason this suite is more than just a series of unit tests, feels more like a set of integration tests since we
  * are testing a chained flow of events.
+ * todo: refactor this whole lot due to resque
  */
-class MongoTripodQueueOperationsTest extends MongoTripodTestBase
+class MongoTripodQueueOperationsDeadTestRefactor
```
robotrobot: Resolve in PR > 30%

test/unit/mongo/MongoTripodQueueTest.php
```diff
@@ -1,9 +1,11 @@
 <?php
 require_once 'MongoTripodTestBase.php';
 require_once 'src/mongo/MongoTripod.class.php';
-require_once 'src/mongo/queue/MongoTripodQueue.class.php';
 
-class MongoTripodQueueTest extends MongoTripodTestBase
+/**
+ * todo: refactor this whole class as queue is dead.
+ */
+class MongoTripodQueueDeadTestRefactor
```
robotrobot: Resolve in PR > 30%

composer.json
```diff
@@ -18,7 +18,8 @@
         }
     ],
     "require": {
-        "semsol/arc2": "v2.2.4"
+        "semsol/arc2": "v2.2.4",
+        "chrisboulton/php-resque": "dev-master#98fde571db008a8b48e73022599d1d1c07d4a7b5"
```
lordtatty: Pinned to a commit rather than 1.2?  I'm not against it, just wondering the rationale in this instance - is there a particular feature/bugfix needed?  If it's primarily to keep things up to date then there are more recent changes to php resque which we could pin to.

src/mongo/ModifiedSubject.class.php
```diff
@@ -46,7 +51,7 @@ public function attach(SplObserver $observer)
      */
     public function detach(SplObserver $observer)
```
lordtatty: Looks like it doesn't need this param anymore.

composer.json
robotrobot: I gave in to peer pressure.

src/mongo/ModifiedSubject.class.php
robotrobot: Indeed. But it's part of the interface, unfortunately :-(

composer.json
astilla: What's the story here then as I hit a similar dilemma on TADC where it was currently pinned to that dev hash but thought I should update it to what Depot used (1.2).

rsinger: 1.2 is nearly 3 years old and the php-resque project party line is to use dev-master.  See: https://github.com/chrisboulton/php-resque/issues/203#issuecomment-88145890

lordtatty: Oh awesome, I'm glad there is some clarity on that now.

scripts/mongo/worker.inc.php
rsinger: Monolog?

src/mongo/delegates/MongoTripodUpdates.class.php
```diff
@@ -706,23 +706,19 @@ public function processSyncOperations(ChangeSet $cs, $deletedSubjects, $contextA
...
-    protected function getQueue()
-    {
-        if(empty($this->queue))
-        {
-            $this->queue = new MongoTripodQueue();
+        $operations = $this->getAsyncOperations();
+        if (!empty($operations)) {
+            $data = array(
+                "changeSet" => $cs->to_json(),
+                "deletedSubjects" => $deletedSubjects,
```
rsinger: Shouldn't we be able to get the deleted subjects from the ChangeSet object?

src/mongo/MongoTripodConstants.php
```diff
@@ -76,3 +77,34 @@
 
 // Mongo constants
 define('DEFAULT_MONGO_CONNECT_TIMEOUT_MS', 20000);
+
+// queue name
+$tDiscoverQueueEnv = getenv("TRIPOD_DISCOVER_QUEUE");
```
rgubby: Add these env vars to the build file so it fails to do an ```ant init``` (or grunt when we move to it) if they aren't set?

src/mongo/MongoTripodConstants.php
```diff
@@ -76,3 +77,34 @@
...
+if (!empty($tDiscoverQueueEnv))
+{
+    define("TRIPOD_DISCOVER_QUEUE",$tDiscoverQueueEnv);
+}
+elseif(defined('APP_ENV'))
+{
+    define("TRIPOD_DISCOVER_QUEUE","tripod::".APP_ENV."::discover");
```
rgubby: constant for "tripod"

rgubby:
When we merge this, can we get some a wiki page created for how this works.

src/mongo/jobs/ApplyOperation.class.php
```diff
@@ -0,0 +1,45 @@
...
+        {
+            $this->debugLog("ApplyOperation::perform() start");
+
+            $timer = new Timer();
+            $timer->start();
+
+            $this->validateArgs();
+
+            // set the config to what is received
+            MongoTripodConfig::setConfig($this->args["tripodConfig"]);
```
rgubby: Not sure if i've missed something, but where is ```$this->args``` coming from / being set ?

rsinger: Resque

src/mongo/MongoTripodConstants.php
rsinger: I think it should fail in tripod if these aren't set, not in the build file.

src/tripod.inc.php
```diff
@@ -1,20 +1,37 @@
...
-    if(!defined('ARC_DIR')) define('ARC_DIR', TRIPOD_DIR. '../vendor/semsol/arc2/' );
-}
 
-require_once ARC_DIR.'ARC2.php';
+if(!defined('TRIPOD_DIR')) define('TRIPOD_DIR', dirname(__FILE__) . '/' );
 
+require_once TRIPOD_DIR.'../vendor/autoload.php';
+
+// use latest brought in via composer
+if(!defined('ARC_DIR')) define('ARC_DIR', TRIPOD_DIR. '../vendor/semsol/arc2/' );
```
rgubby: Delete out ```/lib/arc``` if we're no longer using it

src/mongo/MongoTripodConstants.php
rgubby: woops - sorry, got my projects mixed up - you're right.

PR #55, opened 2015-05-18
90%: Queue bulk composite generation

rsinger:
This refactors the createTables and createViews scripts to allow them to push the actual regeneration to the resque queue.  You may also specify which queue you want these to go on, so as not to clog the regular process queue.

It also introduces a createSearchDocuments script which never existed in Tripod.

The major difference between these and how they work currently is that the current version loops over the CBDs and generates the composites serially.  This version gives the option to loop over the CBDs and submits a job to generate the composite, so the actual generation can be scaled/stopped as needed without starting from the beginning to resume.

If you don't pass the -a/--async flag it works exactly as it does now.

### DIFF
--- a/scripts/mongo/createSearchDocuments.php
+++ b/scripts/mongo/createSearchDocuments.php
@@ -1,7 +1,7 @@
 <?php
 
 $options = getopt(
-    "c:s:q:ht:i:a",
+    "c:s:q:hd:i:a",
     array(
         "config:",
         "storename:",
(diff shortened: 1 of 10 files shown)


scripts/mongo/createSearchDocuments.php
```diff
@@ -0,0 +1,178 @@
...
+php $scriptName -c/--config path/to/tripod-config.json -s/--storename store-name [options]
+
+Options:
+    -h --help               This help
+    -c --config             path to MongoTripodConfig configuration (required)
+    -s --storename          Store to create views for (required)
+    -d --spec               Only create for specified search document specs
+    -i --id                 Resource ID to regenerate search documents for
+    -a --async              Generate table rows via queue
+    -q --queue              Queue name to place jobs on (defaults to configured TRIPOD_APPLY_QUEUE value)
```
robotrobot: If you specify an alternative queue, how do you know anything is listening to it?

rsinger: You don't - the purpose is you'd fire up a worker to listen for that queue to keep it from polluting/clogging the main queue

malcyL:
:+1:

PR #56, opened 2015-05-19
70%: Introduce namespaces

rsinger:
This branch introduces namespaces to Tripod.

Feel free to argue about which classes should go in which namespaces and how they should be named.

The way I've divided them up currently (although this is totally open for discussion) is:

` Tripod `
* ~~` ITripod `~~ ` IDriver `
* ` ITripodStat `
* ` ChangeSet `
* ` ExtendedGraph `
* ` Labeller `
* ` Timer `
* ~~` ITripodSearchProvider ` (perhaps could just be ` iSearchProvider `, thoughts?)~~ ` ISearchProvider ` (formerly ` iTripodSearchProvider `)

` Tripod\Mongo ` (btw, it will take *forever* to get used to this inversion)
* ~~` Tripod `~~ ` Driver ` (formerly ` MongoTripod `)
* ~~` TripodBase `~~ ` DriverBase ` (formerly ` MongoTripodBase `)
* ` MongoGraph `
* ` Labeller ` (formerly ` MongoTripodLabeller `)
* ` ImpactedSubject ` (in current master, this is ` ModifiedSubject `, but that concept has been changed in this feature)
* ` Config ` (formerly ` MongoTripodConfig `)
* ` TriplesUtil `
* ` IndexUtils `
* ` NQuadSerializer ` (formerly ` MongoTripodNQuadSerializer `)
* ` MongoSearchProvider ` (this wasn't renamed because this is defining a Mongo-based search provider)
* ` Updates ` (formerly ` MongoTripodUpdates `)
* ` TransactionLog ` (formerly ` MongoTransactionLog ` - this might have been a mistake, as it's probably more like ` MongoSearchProvider `)
* ` SearchDocuments ` (formerly ` MongoTripodSearchDocuments `)

` Tripod\Mongo\Composites `
* ` IComposite ` (views, tables, and search are now defined as *composites*, this is an interface common among all)
* ` CompositeBase `
* ` Views ` (formerly ` MongoTripodViews `.  ~~Do the composites need their own part of the namespace? I feel that might be overkill~~)
* ` Tables ` (formerly ` MongoTripodTables `)
* ` SearchIndexer ` (formerly ` MongoTripodSearchIndexer `)

` Tripod\Mongo\Jobs `
* ` ApplyOperation ` (this is the background job to generate the composites - ~~the jobs could arguably have their own ` Tripod\Mongo\Jobs ` namespace. Discuss~~ moved into their own package)
* ` DiscoverImpactedSubjects ` (same as above)
[...]

PR #57, opened 2015-05-20
90%: Treat tables, views, and search differently

rsinger:
I found when working on #56 that IndexUtils wasn't properly updated to deal with multiple datasources, which causes failures when it runs across tablespecs, viewspecs, and search document specs.

This branch excludes views, search, and table indexes from the regular index regeneration process and then specifically handles views and table rows (search indexes are done via some other mechanism).

### DIFF
--- a/src/mongo/util/IndexUtils.class.php
+++ b/src/mongo/util/IndexUtils.class.php
@@ -1,31 +1,37 @@
 <?php
 require_once(TRIPOD_DIR."mongo/MongoTripodConfig.class.php");
 
+/**
+ * Class IndexUtils
+ */
 class IndexUtils
 {
     /**
-     * Ensures the index for the given $dbName. As a consequence, sets the global
+     * Ensures the index for the given $storeName. As a consequence, sets the global
      * MongoCursor timeout to -1 for this thread, so use with caution from anything
      * other than a setup script
      * @param bool $reindex - force a reindex of existing data
-     * @param null $dbName - database name to ensure indexes for
+     * @param null $storeName - database name to ensure indexes for
      * @param bool $background - index in the background (default) or lock DB whilst indexing
      */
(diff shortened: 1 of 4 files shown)


src/mongo/util/IndexUtils.class.php
```diff
@@ -33,25 +39,51 @@ public function ensureIndexes($reindex=false,$dbName=null,$background=true)
...
+                    $indexes = array("_id.type"=>1);
+                    if(isset($spec['ensureIndexes']))
+                    {
+                        $indexes = array_merge($indexes, $spec['ensureIndexes']);
+                    }
+                    if ($reindex)
+                    {
+                        $collection->deleteIndexes();
+                    }
+                    $collection->ensureIndex($indexes, array("background"=>$background));
```
robotrobot: We should use `sparse='true'` here to save disk space 

http://php.net/manual/en/mongocollection.ensureindex.php

src/mongo/util/IndexUtils.class.php
```diff
@@ -33,25 +39,51 @@ public function ensureIndexes($reindex=false,$dbName=null,$background=true)
...
```
robotrobot: use sparse

rsinger:
Search indexes in config are defined differently than the other specs and were never covered under this method.

That said, it probably makes sense to force an ensureIndexes to _id.type on the search doc collections just like the others, will do.

src/mongo/util/IndexUtils.class.php
robotrobot: Scratch that, docs say "If a sparse index would result in an incomplete result set for queries and sort operations, MongoDB will not use that index unless a hint() explicitly specifies the index."

rsinger:
Ah, actually, search documents don't define indexes because we only search via regex, so they'd be useless, anyway.  Good reason for not merging tables and search documents, honestly.

PR #65, opened 2015-06-05
50%: Only return spec fields in composite generation

rsinger:
This branch reduces the amount of data returned from Mongo so that it only returns the fields that are defined in the composite specification during the generation of the composites

### DIFF
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1666,6 +1666,94 @@ class Config
         return $fields;
     }
 
+
+    /**
+     * For a given 'level' in a composite specification (views, table rows, search documents), return the relevant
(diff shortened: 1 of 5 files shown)


src/mongo/Config.class.php
```diff
@@ -791,7 +796,14 @@ protected function getDefinedPredicatesInSpecs($storename)
             {
                 continue;
             }
-            $predicates[$spec[_ID_KEY]] = array_unique($this->getDefinedPredicatesInSpecBlock($spec));
+            if(isset($this->specPredicates[$spec[_ID_KEY]]))
```
robotrobot: What's this even doing? Doc..

rsinger: That part isn't even new!

PR #71, opened 2015-07-06
90% - In place updates

robotrobot:
On very large target graphs, the seriali(s|z)ation inside `Updates::applyChangeSet()` gave an unacceptable performance overhead, especially if just one or two triples were being updated.

This PR deals with the native mongodoc format in there, chucking the graphs, and also implementing in-place updates whilst we're there for good measure


src/mongo/delegates/Updates.class.php
```diff
@@ -467,10 +468,10 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
 
-                $targetGraph = new MongoGraph();
-                $targetGraph->add_tripod_array($doc);
+//                $targetGraph = new MongoGraph();
+//                $targetGraph->add_tripod_array($doc);
 
-                if(!$targetGraph->has_triples_about($subjectOfChange))
+                if(empty($doc))
                 {
                     $upsert = true;
```
rsinger: I realize this isn't new here, but I think this is a bug: this could potentially set ` upsert=true ` for every change, couldn't it?

robotrobot: How so? Perhaps I have missed something... If the document for update is empty it is an insert surely?

rsinger: Yes, but ` $doc ` for the next ` $change ` might *not* be empty - we never set ` $upsert ` back to false, so if the first one is new, it will treat all the other documents as new.

rsinger:
In the interest of fixing broken windows, I think we should consider changing line 440 of Updates.class.php from
```php
        if (preg_match('/^CBD_/',$this->getCollection()->getName()))
```
to
```php
    if(in_array($this->getCollection()->getName(), $this->getConfigInstance()->getPods($this->getStoreName())))
```
We don't actually enforce this notion of 'podname must be prefixed with CBD_', and I'm not sure should, anway.

src/mongo/delegates/Updates.class.php
```diff
@@ -659,6 +668,75 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
         }
     }
 
+    private function getAdditionsRemovalsGroupedByNsPredicate($cs,$additions,$removals)
```
rsinger: More 70% feedback here, but some type hinting on these.  In the interest of this code review, I'm assuming it's ChangeSet, Array, Array

src/mongo/delegates/Updates.class.php
```diff
@@ -659,6 +668,75 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
+    {
+        $additionsGroupedByNsPredicate = $this->getChangesGroupedByNsPredicate($cs,$additions);
+        $removalsGroupedByNsPredicate = $this->getChangesGroupedByNsPredicate($cs,$removals);
+
+        $mergedResult = array();
+        foreach($additionsGroupedByNsPredicate as $predicate => $values)
+        {
+            if (!isset($mergedResult[$predicate]))
```
rsinger: I don't think this is necessary -  you can assume that it's not set at this point.  It is necessary for the removals, however.

src/mongo/delegates/Updates.class.php
```diff
@@ -478,61 +479,87 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
-                    $object = $cs->get_subject_property_values($r["value"],$this->labeller->qname_to_uri("rdf:object"));
-
-                    $isUri = ($object[0]['type']=="uri");
-
-                    if ($isUri)
+                    if (count($valueObject)>0)
                     {
-                        $targetGraph->add_resource_triple($subjectOfChange,$predicate,$object[0]["value"]);
+                        // unique value object
+                        $valueObject = array_map("unserialize", array_unique(array_map("serialize", $valueObject)));
```
rsinger: What's the purpose of the serialization/unserialization?

src/mongo/delegates/Updates.class.php
```diff
@@ -478,61 +479,87 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
+                    // remove
+                    if (isset($additionsRemovals["removals"]))
                     {
-                        $subjectsAndPredicatesOfChange[$subjectOfChange][] = $predicate;
+                        $elemsToRemove = array();
+                        foreach ($additionsRemovals["removals"] as $removal)
+                        {
+                            foreach($removal as $valueType => $v) { // should only have one k/v
+                                $found = false;
+                                for ($i=0; $i<count($valueObject);$i++) {
```
rsinger: I think all of this could be simplified to:
```php
foreach($removal as $r)
{
    $valueIndex = array_search($r, $valueObject);
    if($valueIndex === false) 
    {
        $v = array_pop(array_values($r));
        $this->errorLog("Removal value {$subjectOfChange} {$predicate} {$v} does not appear in target document to be updated",array("doc"=>$doc));
        throw new \Exception("Removal value {$subjectOfChange} {$predicate} {$v} does not appear in target document to be updated");
    }
    else
    {
        $elemsToRemove[] = $valueIndex;
    }
}

File: src/mongo/delegates/Updates.class.php [hunk 6]
```diff
@@ -478,61 +479,87 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
                     {
-                        $targetGraph->add_resource_triple($subjectOfChange,$predicate,$object[0]["value"]);
+                        // unique value object
+                        $valueObject = array_map("unserialize", array_unique(array_map("serialize", $valueObject)));
+
+                        if (count($valueObject)==1)
+                        {
+                            $valueObject = $valueObject[0]; // un-array if only one value
+                        }
+                        $this->addOperatorToChange($mongoUpdateOperations,'$set',array($nsPredicate=>$valueObject));
```
rsinger: I kind of feel we should make these operations available as constants.  ` "$set" ` and ` '$set' ` seem error prone to me and we can pretty easily avoid this with something like ` OPERATION_SET `.

src/mongo/delegates/Updates.class.php
```diff
@@ -541,42 +568,22 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
-                    if($_new_version == 0) {
-                        $newDocument[_CREATED_TS] = new \MongoDate();
-                    } else {
-                        if(isset($doc[_CREATED_TS])) {
-                            $newDocument[_CREATED_TS] = $doc[_CREATED_TS];
-                        }
-                    }
-                    array_push($newCBDs, $newDocument);
-                    $updates[] = array("criteria"=>$criteria,"change"=>$newDocument);
+                    $this->addOperatorToChange($mongoUpdateOperations,'$inc',array(_VERSION=>1));
```
rsinger: shouldn't this be ` array(_VERSION=>($doc[_VERSION] + 1)) `?

src/mongo/delegates/Updates.class.php
```diff
@@ -659,6 +668,75 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
+    {
+        if (!isset($changes[$operator]) || !is_array($changes[$operator]))
+        {
+            $changes[$operator] = array();
+        }
+        foreach($kvp as $key=>$value)
+        {
+            if (isset($changes[$operator][$key]))
+            {
+                $value = array_merge($value,$changes[$operator][$key]);
```
rsinger: I have no idea if this matters, but this will put ` $value ` as the first value of the array.  Would that possibly affect methods, such as get_first_resource/get_first_literal?

src/mongo/delegates/Updates.class.php
```diff
@@ -585,13 +592,14 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
...
                     "findAndModify" => $this->getCollection()->getName(),
                     "query" => $update['criteria'],
-                    "update" => $update['change'],
+                    "update" => $update['changes'],
                     "upsert" => $upsert,
                     "new"=>true
                 );
 
                 try{
                     $result = $this->getDatabase()->command($command);
```
rsinger: I think we should use ` MongoCollection::findAndModify ` rather than ` MongoDB::command ` for this.

src/mongo/delegates/Updates.class.php
```diff
@@ -622,6 +630,7 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
                 }
             }
 
+            // todo: work out what happens with these
```
rsinger: Do you have a hand-wavy idea of how we'll do this?

src/mongo/delegates/Updates.class.php
robotrobot: Got it

src/mongo/delegates/Updates.class.php
robotrobot: http://stackoverflow.com/questions/307674/how-to-remove-duplicate-values-from-a-multi-dimensional-array-in-php

```
[
  ["u"=>"foo:bar"],
  ["u"=>"foo:bar"],
  ["u"=>"foo:baz"]
]
```

to 

```
[
  ["u"=>"foo:bar"],
  ["u"=>"foo:baz"]
]
```

src/mongo/delegates/Updates.class.php
robotrobot: There is never any ordering guarantee on those methods nor should calling code rely a certain value being returned first.

src/mongo/delegates/Updates.class.php
robotrobot: mebbe

src/mongo/delegates/Updates.class.php
robotrobot: not really. It was easy before because you have an empty doc. Will do by 50% PR

src/mongo/delegates/Updates.class.php
```diff
@@ -467,10 +468,10 @@ protected function applyChangeSet(\Tripod\ChangeSet $cs, $originalCBDs, $context
 
                 $doc = $this->getDocumentForUpdate($subjectOfChange, $contextAlias, $originalCBDs);
 
-                $targetGraph = new MongoGraph();
-                $targetGraph->add_tripod_array($doc);
+//                $targetGraph = new MongoGraph();
+//                $targetGraph->add_tripod_array($doc);
```
robotrobot: remove comment

src/mongo/delegates/Updates.class.php
robotrobot: +1

src/mongo/delegates/Updates.class.php
rsinger: Ah, neat.  Will add that to the bag of tricks.

src/mongo/delegates/Updates.class.php
rsinger: Yep, that's reasonable.

src/mongo/delegates/Updates.class.php
rsinger: They effectively do the same thing, MongoCollection::findAndModify() just seems a little nicer, idiomatically.

rsinger:
Holy smokes!

src/mongo/Config.class.php
```diff
@@ -2079,4 +2089,23 @@ private static function getenv($env, $default = false)
...
+
+    /**
+     * @static
+     * @return \Psr\Log\LoggerInterface;
+     */
+    public static function getLogger()
+    {
+        if (self::$logger == null)
+        {
+            $log = new \Monolog\Logger('TRIPOD');
```
rsinger: Not critical, but should this be configurable via envvar or something?

src/mongo/delegates/Updates.class.php
```diff
@@ -865,7 +901,7 @@ protected function lockAllDocuments($subjectsOfChange, $transaction_id, $context
             }else{
 
                 if(count($lockedSubjects)) //If any subject was locked, unlock it
-                $this->unlockAllDocuments($transaction_id);
+                    $this->unlockAllDocuments($transaction_id);
```
rsinger: Can we wrap this conditional in curly braces?  These kind of ifs make me a little uncomfortable if they're not explicitly one-liners.

test/unit/mongo/MongoTripodDriverTest.php
```diff
@@ -91,9 +91,10 @@ public function testSelectSingleValue()
         $this->assertEquals($expectedResult,$actualResult);
     }
 
-    public function testDescribe()
+    public function testGraph()
     {
-        $expectedResult =
+        $expectedResult = new \Tripod\ExtendedGraph();
+        $expectedResult->add_turtle(
             "<http://talisaspire.com/resources/3SplCtWGPqEyXcDiyhHQpA> <http://purl.org/dc/terms/isVersionOf> <http://talisaspire.com/works/4d101f63c10a6> .
```
rsinger: Eventually we should probably move these nearly 90 lines of ntriples (!!) into their own fixture files, but not on this day.

rsinger:
:+1:

rsinger:
A much needed change!

PR #79, opened 2015-07-09
90% - Detects/Retries a failed request to resque

robotrobot:
We can't rely on `php-resque` to always throw exceptions on its own so we also do a hack to check the status to make sure the job is def in the queue.

Backs off for 1s with 5 retries if a job isn't added to the queue.

### DIFF
--- a/src/exceptions/JobException.class.php
+++ b/src/exceptions/JobException.class.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Tripod\Exceptions;
+
+// @codeCoverageIgnoreStart
+
+/**
+ * Class SearchException
+ * @package Tripod
+ */
+class JobException extends Exception {
+
+} // @codeCoverageIgnoreEnd
--- a/src/mongo/base/JobBase.class.php
+++ b/src/mongo/base/JobBase.class.php
@@ -1,6 +1,9 @@
 <?php
 
 namespace Tripod\Mongo\Jobs;
+use Tripod\Exceptions\Exception;
+use Tripod\Exceptions\JobException;
+
 /**
  * Todo: How to inject correct stat class... :-S
  */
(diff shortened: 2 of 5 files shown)


test/unit/mongo/ApplyOperationTest.php
```diff
@@ -252,6 +252,78 @@ public function testCreateJobDefaultQueue()
...
+            ->setMethods(array('enqueue','getJobStatus','warningLog'))
+            ->getMock();
+
+        $applyOperation->expects($this->any())->method("enqueue")->will($this->returnValue("sometoken"));
+        $applyOperation->expects($this->any())->method("getJobStatus")->will($this->returnValue(false));
+
+        // expect 5 retries. Catch this with call to warning log
+        $applyOperation->expects($this->exactly(5))->method("warningLog");
+
+        $exceptionThrown = true;
```
kiyanwang: shouldnt this be initialised to false?

test/unit/mongo/ApplyOperationTest.php
```diff
@@ -252,6 +252,78 @@ public function testCreateJobDefaultQueue()
...
+            ->setMethods(array('warningLog','enqueue'))
+            ->getMock();
+
+        $e = new Exception("Connection to Redis failed after 1 failures.Last Error : (0) php_network_getaddresses: getaddrinfo failed: nodename nor servname provided, or not known");
+        $applyOperation->expects($this->any())->method("enqueue")->will($this->throwException($e));
+
+        // expect 5 retries. Catch this with call to warning log
+        $applyOperation->expects($this->exactly(5))->method("warningLog");
+
+        $exceptionThrown = true;
```
kiyanwang: shouldnt this be initialised to false?
the catch statemement below also sets this to true, so when/how will it ever be false?

kiyanwang:
approach looks ok ,  just those two issues with the tests.

rsinger:
(0 to 3 redis queries/save)

src/exceptions/JobException.class.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Tripod\Exceptions;
+
+// @codeCoverageIgnoreStart
+require_once TRIPOD_DIR . '/exceptions/Exception.class.php';
```
rsinger: I assume this was added out of necessity, can you explain what was going :boom: to add this?  I've been trying to move Tripod away from these explicitly require_once's

src/mongo/base/JobBase.class.php
```diff
@@ -76,10 +79,64 @@ public function errorLog($message, $params = null)
      * @param string $queueName
      * @param string $class
      * @param array $data
+     * @retry int $retryAttempts if queue fails, retry x times before throwing an exception
```
rsinger: This param doc isn't right

src/mongo/base/JobBase.class.php
```diff
@@ -76,10 +79,64 @@ public function errorLog($message, $params = null)
...
+            $token = $this->enqueue($queueName, $class, $data);
+            if(!$this->getJobStatus($token))
+            {
+                $this->errorLog("Could not retrieve status for queued $class job - job $token failed to $queueName");
+                throw new \Exception("Could not retrieve status for queued job - job $token failed to $queueName");
+            }
+            else
+            {
+                $this->debugLog("Queued $class job with $token to $queueName");
+                return $token;
```
rsinger: :+1: to this - it was a pretty big oversight that it was left out before.

src/exceptions/JobException.class.php
robotrobot: I copied all the existing exceptions!

rsinger:
Actually it should have gone from 1 to 2: it should only have affected the DiscoverImpactedSubjects job

src/exceptions/JobException.class.php
rsinger: no-words-homer-into-brush

robotrobot: LOL

Issue #81, opened 2015-07-19
A null value can be added to a graph, but then throws an exception when removing


lordtatty:
Currently tripod can add null values, but cannot remove them, resulting in an exception.
```
$emptyGraph = new \Tripod\ExtendedGraph()'
$testGraph = new \Tripod\ExtendedGraph();
$testGraph->add_literal_triple('http://testme.fake', 'http://purl.org/ontology/bibo/Bill', null);

// Works - saves the new triple to the graph with a null value
$tripod->saveChanges($emptyGraph, $testGraph);

// Fails - cannot remove the triple, throws exception "Removal value http://testme.fake http://purl.org/ontology/bibo/Bill does not appear in target document to be updated"
$tripod->saveChanges($testGraph, $emptyGraph);
```

The error is thrown here: https://github.com/talis/tripod-php/blob/master/src/mongo/delegates/Updates.class.php#L515

``$valueIndex = array_search($removal, $valueObject);`` is returning null, because of line https://github.com/talis/tripod-php/blob/master/src/mongo/delegates/Updates.class.php#L490.  The isset will return false against a null value, meaning tripod assumes the value to be an array of values already, so does not wrap the single null value in an array, meaning the array structure will be wrong when trying to match against the array_search.  Resulting false, and throwing the exception.

1.  Should tripod allow null values to be saved to a graph?  I feel like it should be throwing an exception if trying to add anything other than a string.
2.  We need to modify tripod to be able to cope with any graphs which already have null values inserted.

rsinger:
Related to #47

rsinger:
FWIW, this is a bug with ExtendedGraph - It's what should be preventing null values.

robotrobot:
Yup, agree @rsinger . ExtendedGraph should guard against adding null triples in the first place (should probably just silently continue if you try rather than save a null to the internal array).

@lordtatty if that was the case I am guessing your code above would work because `$testGraph` wouldn't actually contain anything, i.e. it would be functionally equivalent to `$emptyGraph`

lordtatty:
Agree that this is the way tripod should be working, but what about handling existing data which has already been set as null?  If we prevent extended graph from accepting a null value then any null values will be orphaned, as there will be no way to delete them (because there will be no way to have a 'before' graph with a null value in it).  This could also have knock-on effects for anything which is relying on those triples existing.  

[...]

PR #82, opened 2015-07-21
90% - Prevent null values being added to ExtendedGraph

lordtatty:
To address issue https://github.com/talis/tripod-php/issues/81

This will ensure that subjects, predicates, and resource values are always strings.  Literal values can be any scalar - we would like to enforce strings for literal values too, but doing so may break legacy databases which use non-strings as values.

### DIFF
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -174,6 +174,7 @@ class ExtendedGraph
      * @param string $s
      * @param string $p
      * @param array $o_info
+     * @throws Exceptions\Exception
      * @return bool
      */
     private function _add_triple($s, $p, Array $o_info) {
@@ -186,19 +187,19 @@ class ExtendedGraph
         if(!$this->isValidResourceValue($p)){
             throw new \Tripod\Exceptions\Exception("The predicate is invalid");
         }
-            if (!isset($this->_index[$s])) {
-                $this->_index[$s] = array();
-                $this->_index[$s][$p] = array($o_info);
+        if (!isset($this->_index[$s])) {
+            $this->_index[$s] = array();
+            $this->_index[$s][$p] = array($o_info);
+            return true;
+        } elseif (!isset($this->_index[$s][$p])) {
+            $this->_index[$s][$p] = array($o_info);
+            return true;
+        } else {
+            if (!in_array($o_info, $this->_index[$s][$p])) {
+                $this->_index[$s][$p][] = $o_info;
                 return true;
-            } elseif (!isset($this->_index[$s][$p])) {
-                $this->_index[$s][$p] = array($o_info);
-                return true;
-            } else {
-                if (!in_array($o_info, $this->_index[$s][$p])) {
-                    $this->_index[$s][$p][] = $o_info;
-                    return true;
-                }
             }
+        }
         return false;
     }
 
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -141,7 +141,10 @@ class ExtendedGraph
      * @return boolean true if the triple was new, false if it already existed in the graph
      */
     public function add_resource_triple($s, $p, $o) {
-        return $this->_add_triple($s, $p, array('type' => strpos($o, '_:' ) === 0 ? 'bnode' : 'uri', 'value' => $o));
+        if($this->isValidResourceValue($o)) {
+            return $this->_add_triple($s, $p, array('type' => strpos($o, '_:') === 0 ? 'bnode' : 'uri', 'value' => $o));
+        }
+        return false;
     }
 
     /**
@@ -154,14 +157,17 @@ class ExtendedGraph
      * @return boolean true if the triple was new, false if it already existed in the graph
      */
     public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
-        $o_info = array('type' => 'literal', 'value' => $o);
-        if ( $lang != null ) {
-            $o_info['lang'] = $lang;
+        if($this->isValidLiteralValue($o)) {
+            $o_info = array('type' => 'literal', 'value' => $o);
+            if ($lang != null) {
+                $o_info['lang'] = $lang;
+            }
+            if ($dt != null) {
+                $o_info['datatype'] = $dt;
+            }
+            return $this->_add_triple($s, $p, $o_info);
         }
-        if ( $dt != null ) {
-            $o_info['datatype'] = $dt;
-        }
-        return $this->_add_triple($s, $p, $o_info);
+        return false;
     }
 
     /**
@@ -171,19 +177,17 @@ class ExtendedGraph
      * @return bool
      */
     private function _add_triple($s, $p, Array $o_info) {
-        if($this->isValidTripleValue($o_info['value'])) {
-            if (!isset($this->_index[$s])) {
-                $this->_index[$s] = array();
-                $this->_index[$s][$p] = array($o_info);
+        if (!isset($this->_index[$s])) {
+            $this->_index[$s] = array();
+            $this->_index[$s][$p] = array($o_info);
+            return true;
+        } elseif (!isset($this->_index[$s][$p])) {
+            $this->_index[$s][$p] = array($o_info);
+            return true;
+        } else {
+            if (!in_array($o_info, $this->_index[$s][$p])) {
+                $this->_index[$s][$p][] = $o_info;
                 return true;
-            } elseif (!isset($this->_index[$s][$p])) {
-                $this->_index[$s][$p] = array($o_info);
-                return true;
-            } else {
-                if (!in_array($o_info, $this->_index[$s][$p])) {
-                    $this->_index[$s][$p][] = $o_info;
-                    return true;
-                }
             }
         }
         return false;
@@ -193,10 +197,26 @@ class ExtendedGraph
     /**
      * Check if a triple value is valid.
      *
+     * Ideally a valid literal value should be a string
+     * but accepting scalars so we can handle legacy data
+     * which was not type-checked.
+     *
      * @return bool
      */
-    protected function isValidTripleValue($value){
-        if(!is_scalar($value) && !is_array($value)){
+    protected function isValidLiteralValue($value){
+        if(!is_scalar($value)){
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Check if a triple value is valid.
+     *
+     * @return bool
+     */
+    protected function isValidResourceValue($value){
+        if(!is_string($value)){
             return false;
         }
         return true;
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -193,13 +193,10 @@ class ExtendedGraph
     /**
      * Check if a triple value is valid.
      *
-     * Currently null values are not valid,
-     * although this may expand to include any non-strings
-     *
      * @return bool
      */
     protected function isValidTripleValue($value){
-        if(!is_string($value) && !is_int($value) && !is_array($value) && !is_float($value) && !is_bool($value)){
+        if(!is_scalar($value) && !is_array($value)){
             return false;
         }
         return true;
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -177,6 +177,10 @@ class ExtendedGraph
      * @return bool
      */
     private function _add_triple($s, $p, Array $o_info) {
+        // Make sure the subject is not an empty string
+        if($s === ""){
+            throw new \Tripod\Exceptions\Exception("The subject cannot be an empty string");
+        }
         // The value $o should already have been validated by this point
         // It's validation differs depending on whether it is a literal or resource
         // So just check the subject and predicate here...
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -141,7 +141,7 @@ class ExtendedGraph
      * @return boolean true if the triple was new, false if it already existed in the graph
      */
     public function add_resource_triple($s, $p, $o) {
-        if($this->isValidTripleValue($s) && $this->isValidTripleValue($p) && $this->isValidTripleValue($o)) {
+        if($this->isValidResourceValue($o)) {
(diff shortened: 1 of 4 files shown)


src/classes/ExtendedGraph.class.php
```diff
@@ -171,24 +171,40 @@ public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
...
+
+    /**
+     * Check if a triple value is valid.
+     *
+     * Currently null values are not valid,
+     * although this may expand to include any non-strings
+     *
+     * @return bool
+     */
+    protected function isValueValid($value){
```
robotrobot: change name to `isValidLiteralValue()` ?

lordtatty: Wouldn't we want to check resource values as well though?  ``isValidTripleValue``?

rsinger: From Tripod's perspective, don't we just need to ensure that subject, predicate, and object are all strings?

rsinger: non-empty strings

lordtatty: Agree, but enforcing strings as opposed to null could be very painful for any graphs which are already storing non-string values (eg. integers).  This current PR to ignore null values could well cause enough headaches without taking out integers at the same time.   I do think we should ultimately enforce strings-only, but maybe one step at a time, once we are comfortable with no nulls.

rsinger: Tripod *only* deals in strings, does it not?

lordtatty: It should only deal with strings, but up until now there has been nothing to enforce it...  Clearly this method will be the right place to enforce that - but is it what we want to do right now?

rsinger: Ok, so strings and integers.

rsinger: Or arrays of strings and integers.

rsinger: I guess what I'm saying is that we should be doing type checking here so we're not just allowing some other bad data that is equally as incompatible as a null value.

lordtatty: How do we deal with legacy data, though?  We have no idea what crazy things have ended up in the graph.  The php data types are:

```
String.
Integer.
Float (floating point numbers - also called double)
Boolean.
Array.
Object.
NULL.
Resource.
```

Integer, floats and booleans could all have ended up in the graph.  Is it possible to add an array via the public ExtendedGraph methods?

Agree that really we should be ignoring anything that isn't a string, but I guess the question is how much we are prepared to break anything currently using tripod to enforce this.

rsinger: Tripod only supports plain literals.  Full-stop.  If other data types are in the documents they would produce invalid RDF, anyway, because we do not support typed literals at all.

src/classes/ExtendedGraph.class.php
```diff
@@ -171,24 +171,40 @@ public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
...
+    /**
+     * Check if a triple value is valid.
+     *
+     * Currently null values are not valid,
+     * although this may expand to include any non-strings
+     *
+     * @return bool
+     */
+    protected function isValidTripleValue($value){
+        if(!is_string($value) && !is_int($value) && !is_array($value) && !is_float($value) && !is_bool($value)){
```
rsinger: Why not ` is_scalar ` here?

lordtatty: Uh yep, I meant to use that, no idea why I didn't.... one moment.

lordtatty:
@rsinger @robotrobot Something I just noticed is https://github.com/talis/tripod-php/blob/master/src/classes/ExtendedGraph.class.php#L141 - ``@return boolean true if the triple was new, false if it already existed in the graph``.

What should be return of ``add_resource_triple`` and ``add_literal_triple `` be if we aren't adding the triple due to the value being invalid?  Null?  False?  At the moment it is false.

lordtatty:
@rsinger @robotrobot also at the moment we are allowing scalars for both literals and resources. Should we restrict resources to only strings? I really can't imagine how having anything other than a string can be legitimate for a uri.

src/mongo/MongoGraph.class.php
```diff
@@ -138,39 +142,65 @@ private function add_tarray_to_index($tarray)
 
     /**
      * Convert from Tripod value object format (comapct) to ExtendedGraph format (verbose)
+     *
      * @param array $mongoValueObject
-     * @return array
+     * @return array| false an array of values or false if the value is not valid
```
rsinger: `array| false` isn't a valid phpdoc, it needs to be `array|bool`

rsinger:
This looks good.  Should we validate `$s` and `$p` while we're here?  Is it possible to add invalid values for them currently?

lordtatty:
@rsinger  I feel like maybe that should be raised as a separate issue so we can get this branch down.

rsinger:
Why can't you just run $s and $p through isValidResourceProperty?

rsinger:
I mean, my real question is, can you even add invalid values there?  I don't see why wouldn't fix it here, if so.

lordtatty:
@rsinger my only real reservation to fixing that in this branch is that I'll have to add a whole bunch of tests to cover that, and this needs to get shipped today, ideally.  It's currently passing as it is so I think the right strategy is to merge this down and fix up the $s and $p in another branch...

rsinger:
Can you just confirm whether or not it's even an issue?  There is no point in making a release to prevent us from adding duff data if we've still left it open to add duff data.

lordtatty:
Added checks for subjects and predicates

lordtatty:
This will now also throw an exception if trying to add a triple with a blank subject

src/classes/ExtendedGraph.class.php
```diff
@@ -171,24 +177,60 @@ public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
...
+        }
+        return true;
+    }
+
+    /**
+     * Check if a triple value is valid.
+     *
+     * @return bool
+     */
+    protected function isValidTripleValue($value){
```
rsinger: Triple value?

lordtatty: I thought that would be a good generic name to cover subject, predicate and value.

rsinger: I think it made more sense when it was named resourcevalue

lordtatty: It's used for more than just validating a resource value though - the only thing it doesn't validate now is a literal value.  ``isValidTripleComponentValue``?

lordtatty: Should I be breaking this out into ``isValidSubject``, ``isValidPredicate``, and ``isValidVaue``?  It might make sense since we are now ensuring that the subject is not an empty string, as we can just add the check into there...

rsinger: No, they're all just uris.

src/classes/ExtendedGraph.class.php
```diff
@@ -154,14 +157,17 @@ public function add_resource_triple($s, $p, $o) {
      * @return boolean true if the triple was new, false if it already existed in the graph
      */
     public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
-        $o_info = array('type' => 'literal', 'value' => $o);
-        if ( $lang != null ) {
-            $o_info['lang'] = $lang;
-        }
-        if ( $dt != null ) {
-            $o_info['datatype'] = $dt;
+        if($this->isValidTripleValue($s) && $this->isValidTripleValue($p) && $this->isValidLiteralValue($o)) {
```
robotrobot: You're doing this already inside `_add_triple()`. Don't do it tice.

src/classes/ExtendedGraph.class.php
```diff
@@ -141,7 +141,10 @@ public function make_resource_array($resource) {
      * @return boolean true if the triple was new, false if it already existed in the graph
      */
     public function add_resource_triple($s, $p, $o) {
-        return $this->_add_triple($s, $p, array('type' => strpos($o, '_:' ) === 0 ? 'bnode' : 'uri', 'value' => $o));
+        if($this->isValidTripleValue($s) && $this->isValidTripleValue($p) && $this->isValidTripleValue($o)) {
```
robotrobot: You're doing this already inside `_add_triple`. Don't do it twice.

src/classes/ExtendedGraph.class.php
```diff
@@ -171,24 +177,60 @@ public function add_literal_triple($s, $p, $o, $lang = null, $dt = null) {
...
-        }
-        elseif (!isset($this->_index[$s][$p])) {
-            $this->_index[$s][$p] = array( $o_info);
-            return true;
-        }
-        else {
-            if ( ! in_array( $o_info, $this->_index[$s][$p] ) ) {
-                $this->_index[$s][$p][] = $o_info;
+        // Make sure the subject is not an empty string
+        if($s === ""){
```
robotrobot: Why do this here? Your validation should be inside an isValid style method. Also predicates are not allowed to be empty either.

robotrobot: isEmpty() is probably better in any case

lordtatty: I was just mooting that with Ross in a comment further down. Wondering if I should break up ``isValidTripleValue`` into more specific validation methods for the subject, predicate and value.

lordtatty: Will do that shortly.

src/classes/ExtendedGraph.class.php
robotrobot: Can we have two methods please. `isValidResource()` - this covers subjects, predicates or objects that are of type URI. Another method - `isValidLiteral()` covers literals only. 

You are currently doing validation in two places - once in _add_triples() and again in the individual methods - pick one or the other don't do both.

Finally you're null checking outside these methods. Do the null checking inside the `isValidResource`. If the validity check fails for resources or predicates I think that is an Exception personally, but for objects we are saying it is allowed but ignored (to keep the peace).

src/mongo/MongoGraph.class.php
```diff
@@ -128,8 +128,24 @@ private function add_tarray_to_index($tarray)
         {
             if($key[0] != '_')
             {
-                $predicate = $this->qname_to_uri($key);
-                $predObjects[$predicate] = $this->toGraphValueObject($value);
+                // Make sure the predicate is valid
+                if($this->isValidTripleValue($key)){
```
robotrobot: I'm not sure you need this. qname_to_uri will fail if $key is not a string that can be matched to a namespace.

src/mongo/MongoGraph.class.php
```diff
@@ -128,8 +128,24 @@ private function add_tarray_to_index($tarray)
...
+                        $predObjects[$predicate] = $graphValueObject;
+                    }
+                }
+            }
+            else if($key == "_id"){
+                // If the subject is not valid then return
+                if(!isset($value['r']) || !$this->isValidTripleValue($value['r'])){
+                    return;
+                }
+                if($value['r'] === ""){
```
robotrobot: again - why is your validity checking not inside the method validating the value? Instead if `isValidWhatever` returns false then throw your exception.

src/classes/ExtendedGraph.class.php
lordtatty: I see your point, in my head I was separating subjects and uri values unnecessarily here, will fix up.

src/mongo/MongoGraph.class.php
```diff
@@ -138,39 +154,65 @@ private function add_tarray_to_index($tarray)
...
                     $simpleGraphValueObject[] = array(
-                        'type'=>($type==VALUE_LITERAL) ? 'literal' : 'uri',
+                        'type'=>$valueTypeLabel,
                         'value'=>($type==VALUE_URI) ? $this->_labeller->qname_to_alias($value) : $value);
                 }
             }
         }
+        // If we don't have any values, then respond with false
+        if(empty($simpleGraphValueObject)){
+            return false;
```
robotrobot: Why are we returning false here? Instead why not return empty array (as before) and have the calling code do the empty()? 

Returning false to indicate no values out of this method seems very php4. We've unfortunately inherited methods like that from ExtendedGraph (which came from another library) but we don't have to perpetuate the behaviour.

src/classes/ExtendedGraph.class.php
lordtatty: I hadn't spotted that duplicate validation - that was a leftover from a previous iteration.  Removing.

src/mongo/MongoGraph.class.php
lordtatty: That's true, but it will throw an error rather than an exception if the predicate is an empty string.  I've added ``testAddTripodArrayContainingEmptyPredicate`` to assert that a php error is thrown in this circumstance.  It's not a perfect test, but I think it's probably worth it to not have to unnecessarily validate the predicate.

lordtatty:
@robotrobot @rsinger Fixed up from PR comments:  
  - Renamed isValidTripleValue back to isValidResourceValue, and added the empty string check into there.  
  - Removed an unnecessary predicate validation which is being covered by the the labeller, and added appropriate tests to cover that.  
  - toGraphValueObject no longer returns false if the array is empty.

src/classes/ExtendedGraph.class.php
```diff
@@ -189,6 +203,37 @@ private function _add_triple($s, $p, Array $o_info) {
...
+     * Check if a triple value is valid.
+     *
+     * Ideally a valid literal value should be a string
+     * but accepting scalars so we can handle legacy data
+     * which was not type-checked.
+     *
+     * @param string $value
+     * @return bool
+     */
+    protected function isValidLiteralValue($value){
```
robotrobot: Too verbose, rename `isValidLiteral()`,

src/classes/ExtendedGraph.class.php
```diff
@@ -189,6 +203,37 @@ private function _add_triple($s, $p, Array $o_info) {
...
+        return true;
+    }
+
+    /**
+     * Check if a triple value is valid.
+     *
+     * @param string $value
+     * @return bool
+     */
+    protected function isValidResourceValue($value){
```
robotrobot: Too verbose, rename is `isValidResource()`

PR #84, opened 2015-07-23
90% Handle Mongo 2.6's stricter enforcement of index key sizes.

malcyL:
Mongo 2.6 has introduced stricter enforcement of the maximum
index key size. By default an index key > 1024 bytes now throws
an exception. This pull requests changes Table Row generation
to capture this error when thrown. The key is then truncated
and the save retried.

### DIFF
--- a/src/mongo/delegates/Tables.class.php
+++ b/src/mongo/delegates/Tables.class.php
@@ -589,7 +589,7 @@ class Tables extends CompositeBase
                 // Remove temp fields from document
 
                 $generatedRow['value'] = array_diff_key($value, array_flip($this->temporaryFields));
-                $collection->save($generatedRow);
+                $this->truncatingSave($collection, $generatedRow);
             }
         }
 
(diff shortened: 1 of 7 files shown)


malcyL:
This has now gone green on Travis.

I've tried to keep the data for the new unit test in it's own table spec, using a different type of resource - but I've still had to amend the existing tests slightly. Perhaps this could be done more cleanly?

src/mongo/delegates/Tables.class.php
```diff
@@ -603,6 +603,100 @@ public function generateTableRows($tableType,$resource=null,$context=null,$queue
...
+     * @param \MongoCollection $collection
+     * @param array $generatedRow The rows to save.
+     */
+    protected function truncatingSave(\MongoCollection $collection, array $generatedRow)
+    {
+        try
+        {
+            $collection->save($generatedRow);
+        } catch (\Exception $e) {
+            // We only trunacte and retry the save if the \MongoCursorException contains this text.
```
rsinger: Typo

test/unit/mongo/MongoTripodConfigTest.php
```diff
@@ -969,9 +969,10 @@ public function testSearchConfigNotPresent()
     public function testGetAllTypesInSpecifications()
     {
         $types = $this->tripodConfig->getAllTypesInSpecifications("tripod_php_testing");
-        $this->assertEquals(10, count($types), "There should be 10 types based on the configured view, table and search specifications in config.json");
+        $this->assertEquals(11, count($types), "There should be 10 types based on the configured view, table and search specifications in config.json");
```
rsinger: There should be 11 types.

rsinger:
:+1:

PR #86, opened 2015-07-30
90%: Implement cursor timeout overrides for composite bulk delete

rsinger:
We allow the dev to override the default MongoCursor timeout in Config, but we don’t actually use it for the bulk composite deletions.

This rectifies that situation.


rsinger:
Hmm, from a year ago.  I believe the reason it was left open/orphaned was because we ran into issues with different versions of Mongo/php-mongo.  I'm not sure that was ever resolved (or if this is still an issue).

I'll recommend closing it and dealing with the issue it was trying to solve if we see it again.

Issue #87, opened 2015-08-04
Tripod Event Hooks


robotrobot:
Our roadmap has a long standing issue about publishing updates to an Apache Storm spout. 

This feels really specific and I wonder instead if what is needed is some kind of event listener or hook system whereby you can plug in custom code to the tripod lifecycle at some pre-defined points. 

This way a plugin to Apache Storm (or a data pump to a number of other systems) could be created outside of the main library. You could also do very app-specific things that should never be considered for inclusion in the main tripod code line, such as maybe Varnish cache invalidation for web pages when data changes.

You could even imagine a plugin that mutates data being saved before it is handled by tripod, similar to Doctrine's [Timstampable](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/timestampable.md) plugin.

Perhaps this could be achieved by first defining which events are available to listen to in tripod. I'll list some here:

* Driver::saveChanges
* CompositeBase::update
* ???

Next, perhaps we should define an interface for our custom code to implement:

```php
interface EventHook {
    /**
     * This method gets called just before the event happens, and is passed the 
     * same arguments as the function it is called from.
     * Useful for hooks that wish to mutate data before an event
     * @param $args array of arguments
     */
    public function pre(array $args);

    /**
     * This method gets called just after the event happens, and is passed the 
     * same arguments as the function it is called from
     * Useful for hooks that wish to report changes after an event
     * @param $args array of arguments
     */
    public function post(array $args);
}
```

Then, we need a way of configuring hooks. I suggest starting with being able to do this in code against the driver:

```php
$tripod = new Driver(
  "CBD_users",
  "myapp" 
);

$tripod->registerHook(MongoTripodConstants.SAVE_CHANGES,"InvalidateUserVarnishCache");
$tripod->registerHook(MongoTripodConstants.SAVE_CHANGES,"PumpUserChangesToApacheStorm");
```

We might want to think later about doing this in config, maybe the collections area so you can define a hook per collection?

```json
        "CBD_users" : {
            "indexes":{
                "email":{
                    "spec:email.l":1
                }
            },
            "hooks": {
                "saveChanges":["InvalidateUserVarnishCache","PumpUserChangesToApacheStorm"]
            }
        }
```

Some questions:

1. Does this pattern seem reasonable?
2. Should post-event hooks run even if the method throws an exception?
[...]

PR #88, opened 2015-08-05
90% implement event hooks

robotrobot:
As per #87 

An initial sketch of the implementation. There is only one event for now. Slightly different to the issue, args depend on what the event is. They need this to be actually useful.

* [x] Tests with some basic test hooks
* [x] Test misbehaving hooks that throw exceptions
* [ ] Document save changes hook

### DIFF
--- a/src/IDriver.php
+++ b/src/IDriver.php
@@ -8,20 +8,18 @@ namespace Tripod;
  */
 interface IDriver
 {
-    public function graph($query,$includeProperties=array());
-
     /**
-     * @deprecated
-     * @abstract
-     * @param $query
+     * Equivalent to CONSTRUCT
+     * @param array $filter conditions to filter by
+     * @param array $includeProperties only include these predicates, empty array means return all predicates
      * @return mixed
      */
-    public function describe($query);
+    public function graph($filter,$includeProperties=array());
 
     /**
      * Return (DESCRIBE) the concise bound description of a resource
-     * @param $resource string uri resource you'd like to describe
-     * @param null $context string uri of the context, or named graph, you'd like to describe from
+     * @param string $resource uri resource you'd like to describe
+     * @param null|string $context string uri of the context, or named graph, you'd like to describe from
      * @return ExtendedGraph
      */
     public function describeResource($resource,$context=null);
@@ -29,15 +27,15 @@ interface IDriver
     /**
      * Return (DESCRIBE) the concise bound descriptions of a bunch of resources
      * @param array $resources uris of resources you'd like to describe
-     * @param null $context string uri of the context, or named graph, you'd like to describe from
+     * @param null|string $context string uri of the context, or named graph, you'd like to describe from
      * @return ExtendedGraph
      */
     public function describeResources(Array $resources,$context=null);
 
     /**
      * Get a view of a given type for a given resource
-     * @param $resource string uri of the resource you'd like the view for
-     * @param $viewType string type of view
+     * @param string $resource uri of the resource you'd like the view for
+     * @param string $viewType string type of view
      * @return ExtendedGraph
      */
     public function getViewForResource($resource,$viewType);
@@ -45,7 +43,7 @@ interface IDriver
     /**
      * Get views for multiple resources in one graph
      * @param array $resources uris of resources you'd like to describe
-     * @param $viewType string type of view
+     * @param string $viewType type of view
      * @return ExtendedGraph
      */
     public function getViewForResources(Array $resources,$viewType);
@@ -53,13 +51,16 @@ interface IDriver
     /**
      * Get views based on a pattern-match $filter
      * @param array $filter pattern to match to select views
-     * @param $viewType string type of view
+     * @param string $viewType type of view
      * @return ExtendedGraph
      */
     public function getViews(Array $filter,$viewType);
 
     /**
      * Returns the etag of a resource, useful for caching
+     * @param string $resource
+     * @param null|string $context
+     * @return string
      */
     public function getETag($resource,$context=null);
 
@@ -81,7 +82,7 @@ interface IDriver
      * @param array $sortBy
      * @param int $offset
      * @param int $limit
-     * @return mixed
+     * @return array
      */
     public function getTableRows($tableType,$filter=array(),$sortBy=array(),$offset=0,$limit=10);
 
@@ -90,7 +91,7 @@ interface IDriver
      * @param $tableType
      * @param $fieldName
      * @param array $filter
-     * @return mixed
+     * @return array
      */
     public function getDistinctTableColumnValues($tableType, $fieldName, array $filter = array());
 
@@ -98,8 +99,8 @@ interface IDriver
     /**
      * Get a count of resources matching the pattern in $query. Optionally group counts by specifying a $groupBy predicate
      * @param $query
-     * @param null $groupBy
-     * @return mixed
+     * @param null|string $groupBy
+     * @return array|int multidimensional array with int values if grouped by, otherwise int
      */
     public function getCount($query,$groupBy=null);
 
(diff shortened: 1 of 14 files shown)


src/IDriver.php
```diff
@@ -19,33 +17,157 @@ public function graph($query,$includeProperties=array());
...
     public function getETag($resource,$context=null);
 
-    // tabular functions
-
-    public function select($query,$fields,$sortBy=null,$limit=null,$context=null);
+    /**
+     * Select data in a tabular format
+     * @param array $filter pattern to match to select views
+     * @param $fields
+     * @param null $sortBy
```
rsinger: Not to trod on your 30%-er with 90%-er comments, but these are really handy and I don't want them to get lost in the actual functionality of this branch: make sure the param types of these aren't null.

src/IDriver.php
```diff
@@ -19,33 +17,157 @@ public function graph($query,$includeProperties=array());
...
+     * @param ExtendedGraph $oldGraph
+     * @param ExtendedGraph $newGraph
+     * @param null $context
+     * @param null $description
+     * @return mixed
+     */
     public function saveChanges(ExtendedGraph $oldGraph, ExtendedGraph $newGraph,$context=null,$description=null);
 
+    /**
+     * Register an event hook, which
```
rsinger: which... what?  DON'T LEAVE US HANGING!!

src/IEventHook.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * This method gets called after the event has successfully completed. The arguments passed depend on the event in
+     * question, see the documentation for that event type for details
+     * If the event throws an exception or fatal error, this method will not be called.
+     * @param $args array of arguments
+     */
+    public function post(array $args);
+
+    /**
+     * This method gets called if the event failed for any reason. The arguments passed should be the same as IEventHook::pre
```
rsinger: Same as `pre` or `post`?

src/mongo/Driver.class.php
```diff
@@ -599,6 +602,24 @@ public function replayTransactionLog($fromDate = null, $toDate = null)
...
+     * @return mixed
+     */
+    public function registerHook($eventType, IEventHook $hook)
+    {
+        switch ($eventType) {
+            case IEventHook::EVENT_SAVE_CHANGES:
+                $this->getDataUpdater()->registerSaveChangesEventHook($hook);
+                break;
+            default:
+                throw new Exception("Unrecognised type $eventType whilst registering event hook");
```
rsinger: Not sure if we want a type-specific exception for this.

src/mongo/base/DriverBase.class.php
```diff
@@ -376,6 +378,38 @@ protected function getCollection()
...
+        foreach ($hooks as $hook)
+        {
+            try
+            {
+                /* @var $hook IEventHook */
+                $hook->$fn($args);
+            }
+            catch (\Exception $e)
+            {
+                // don't let rabbid hooks stop tripod
```
rsinger: rabid.

rabbid = 
rabbid

src/mongo/delegates/Updates.class.php
```diff
@@ -144,6 +151,12 @@ public function saveChanges(
         $context=null,
         $description=null)
     {
+        $this->applyPreHooks($this->saveChangesHooks,array(
```
rsinger: I don't have a better alternative, offhand, but this feels kind of awkward.

I like the intention, though.  I kind of wish there was an easier way to register these hooks, but no matter what, I don't know of an elegant way to pass local variables to the event hook without doing it like this.

src/mongo/base/DriverBase.class.php
```diff
@@ -376,6 +378,38 @@ protected function getCollection()
...
+
+    protected function applyPostHooks($hooks,$args=array())
+    {
+        $this->applyHooks("post",$hooks,$args);
+    }
+
+    protected function applyFailureHooks($hooks,$args=array())
+    {
+        $this->applyHooks("failure",$hooks,$args);
+    }
```
lordtatty: It would be good to have IEventHook::TIMING_PRE, TIMING_POST, TIMING_FAILURE, (I'm open as to the naming), so that we can comfortably call applyHooks directly, and do away with these indirect methods.

src/mongo/delegates/Updates.class.php
```diff
@@ -1204,6 +1233,14 @@ public function setTransactionLog(TransactionLog $transactionLog)
...
     }
 
+    /**
+     * Register save changes event hooks
+     * @param IEventHook $hook
+     */
+    public function registerSaveChangesEventHook(IEventHook $hook)
+    {
+        $this->saveChangesHooks[] = $hook;
+    }
```
lordtatty: How about having a generic register hook method to store the hooks in a multidimensional array, that way we won't have to add/manage different register methods for each new hook.

src/mongo/base/DriverBase.class.php
robotrobot: +1

src/mongo/delegates/Updates.class.php
robotrobot: We only have one hook type at the moment. That sounds like an optimisation for the next hook to be added.

Also remember not all hooks will be on the same delegate (this happens to be on Updates). What happens if those hooks are on views/tables?

lordtatty: We are designing this with the intention of adding more hooks though - Driver:: registerHook() has been written with other hook types in mind.  It seems strange to have parts of the code prepared for new hook types, but other parts baked in like this.  If other hooks are on the roadmap then I would favour doing this earlier, rather than having to take it out later when we want to add another hook.

If we are sure we are going to be adding more hooks, and those hooks could be across multiple delegates, then maybe we should be breaking out into its own class which can be instantiated and shared between delegates.

robotrobot: I'm not sure what you mean by *parts of the code prepared for new hook types, but other parts baked in like this*.

The current design is classic delegate pattern, which is used throughout TARL. The interface method allows you to add hooks of any type (`IDriver::registerHook()`) the implementation of IDriver then looks at the hook types and delegates the registration out to the various components that care about them.

That is a simple design for what is today a one-event pony.

Going further and having some kind of central storage for all hooks of any type in my experience is premature optimisation. We don't know how hooks are going to evolve. 

It is likely that the next hook type (View/TableRow created) might happen out of process behind a resque job in which case your centralised multidimensional array on the instance of `Driver` that registered the hook breaks down. This is because the instance of the Driver object where the hook was registered is not available to the code actually executing the hook.

That's why I'm against implementing anything more complex at this time.

src/IEventHook.php
robotrobot: Same as `pre`. You have no idea where this failed or why. Some of the args passed to post are calculated as a result of the event being successful, so by definition they could not be passed to some cases of fail.

robotrobot: In light of `failure`, I am wondering if we should rename `post`->`success`

rsinger:
Part of me wonders if we're creating more work for ourselves simply because this functionality isn't available in PHP 5.3.  Wouldn't it be easier to implement this via something like Go! AOP Framework (https://github.com/goaop/framework) and make PHP 5.4+ our priority (this needs to be our priority *anyway*).

That way we're not rolling our own solution and have a community of support to lean on for this.

rsinger:
This is true, I don't know a lot of the details of Go! AOP or that it's necessarily suited for our exact purpose, it was just an example.  I just want to be sure that we're not rowing our own course when it may not entirely be necessary.

src/mongo/Driver.class.php
robotrobot: It's kinda one you should only see in dev. Can't see decisions being made in client code because you funked up on the hook config.

PR #89, opened 2015-08-06
90% - Remove incorrect char encoding

lordtatty:
This utf8_encode is garbling diacritics for search documents.  While adding tests I also discovered that strtolower while building the search indices was also causing utf8 issues with the indices. 

## Issue 1
The utf8_encode function is only designed to convert from ISO-8859-1 to UTF-8.  If we pass a utf8 string through this method then anything outside of the normal ASCII range will be converted as if it is ISO-8859-1, resulting in garbled strings.  See the top answer here for a better explanation: http://stackoverflow.com/questions/6700275/utf8-encode-function-purpose

If we still care about converting to UTF-8 here then we would need to think about something along the lines of: iconv(mb_detect_encoding($text, mb_detect_order(), true), "UTF-8", $text), as per http://stackoverflow.com/questions/7979567/php-convert-any-string-to-utf-8-without-knowing-the-original-character-set-or.  However, that approach is imperfect.

As we are not doing these kinds of conversions for other saves in tripod, I don't think it is necessary here, so I think the simplest answer is to simply remove this encoding.

## Issue 2

strtolower does not respect non ascii characters.  To fix this we need to use mb_strtolower and pass "UTF-8" as the second param.

### DIFF
--- a/src/mongo/delegates/SearchDocuments.class.php
+++ b/src/mongo/delegates/SearchDocuments.class.php
@@ -262,15 +262,15 @@ class SearchDocuments extends DriverBase
 
                     if(isset($source[$p])){
                         if(isset($source[$p]['u'])){
-                            $values[] = ($isIndex) ? strtolower(trim($source[$p]['u'])) : trim($source[$p]['u']);
+                            $values[] = ($isIndex) ? mb_strtolower(trim($source[$p]['u']), 'UTF-8') : trim($source[$p]['u']);
                         } else if (isset($source[$p]['l'])){
-                            $values[] = ($isIndex) ? strtolower(trim($source[$p]['l'])) : trim($source[$p]['l']);
+                            $values[] = ($isIndex) ? mb_strtolower(trim($source[$p]['l']), 'UTF-8') : trim($source[$p]['l']);
                         } else {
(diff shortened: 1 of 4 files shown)


src/mongo/delegates/SearchDocuments.class.php
```diff
@@ -330,7 +330,7 @@ private function addValuesToTarget($values, $field, &$target)
         
 		if(count($values) > 0){						
 	        for ($i=0; $i<$limit; $i++) {	        	
-	            $v = utf8_encode($values[$i]); //todo: added by CJC
```
rsinger: I don't have a problem with this change, but it does raise the question as to why it was added in the first place.

lordtatty: Agreed.  Although utf8_encode will only work if the input is ISO-8859-1, any other encoding and it will result in strange characters, so I can't understand the rationale behind adding this here.  There doesn't seem to be any other part of tripod - as far as I can see - which is attempting to ensure things are encoded in utf-8, and doing so is very difficult without knowing the source encoding, so I think it's simply down to the calling app to make sure that UTF-8 is passed in.

rsinger:
:+1:

PR #94, opened 2015-09-04
90%: Tripod stats factory

rsinger:
This builds on the work from #70 but takes a somewhat different approach to how Stats objects are passed around.

This maintains the current Tripod pattern of passing a Stats object to Tripod\Mongo\Driver as part of the constructor, then passes enough stats config for the background jobs to instantiate the same stats objects from a factory class.


rsinger:
Some questions that are probably best answered by @robotrobot, which would be relevant to either this PR or #70:
* We currently aren't keeping track of how many subjects are in a given DiscoverImpactedSubjects or ApplyOperation job: this seems like an significant omission, since we'd have no idea if a very long running job was 1 subject or 1,000.
* The StatsD class doesn't keep track of 'store' or 'pod', so we have no granularity if some databases/collections are incredibly slow or fast or whatever.
* Do we want to track the timing of each individual operation by operation type (e.g. tables, view, search) in ApplyOperation?

src/mongo/Driver.class.php
```diff
@@ -31,19 +31,6 @@ class Driver extends DriverBase implements \Tripod\IDriver
      */
     private $search_indexer = null;
 
-
-    /**
-     * @var array The original read preference gets stored here
-     *            when changing for a write.
-     */
-    private $originalCollectionReadPreference = array();
```
robotrobot: Why have these been removed?

rsinger: That's from #70

rsinger: This branch was branched off of it.

rsinger: I don't actually see any indication that they were referenced anywhere.  It looks like it might have been intended to be what the same properties on Updates.class.php became.

src/mongo/delegates/Views.class.php
```diff
@@ -482,7 +482,7 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
                         array($viewId)
                     );
 
-                    $this->getApplyOperation()->createJob(array($subject), $queueName);
+                    $this->getApplyOperation()->createJob(array($subject), $queueName); // todo: how to pass statsD?
```
robotrobot: todo

src/mongo/Driver.class.php
robotrobot: Ah ok

rsinger:
Added a lot of new stats calls: I noticed that we really only called `increment()` when operations failed (since I guess timing doesn't really mean much, in that case), but I feel that might be an oversight.  It seems like it's as important to record the frequency of an operation as it does the length of time to perform it.  It's possible that this is unnecessary, i.e. maybe graphite already does this for us simply with the timer stats, but this is definitely something I'd need to defer to @robotrobot and @foomatty for guidance.

I also added a 'custom' method ITripodStat which allows the dev to keep stats besides counts and timers.  StatsD has 'gauges', so for that class, this custom method just uses that.  With this, I've added stats for how many subjects are in an ApplyOperation job and how many impacted subjects are found in a DiscoverImpactedSubjects job.

I don't really know at what threshold these stats become noise (or problematic to graphite), but I figure I can add a lot in the PR (where we can see them all) and rip them out after discussion if need be.

rsinger:
Ah, I noticed that for StatsD->timer(), we're also sending an increment (`"1|c","$duration|ms"`), so I guess those added increments are unnecessary.

src/mongo/MongoTripodConstants.php
```diff
@@ -72,11 +74,25 @@
...
+define('MONGO_QUEUE_APPLY_OPERATION','MONGO_QUEUE_APPLY_OPERATION');
+define('MONGO_QUEUE_APPLY_OPERATION_JOB','MONGO_QUEUE_APPLY_OPERATION_JOB');
+define('MONGO_QUEUE_APPLY_OPERATION_SUBJECT','MONGO_QUEUE_APPLY_OPERATION_SUBJECT');
 define('MONGO_QUEUE_APPLY_OPERATION_SUCCESS','MONGO_QUEUE_APPLY_OPERATION_SUCCESS');
+define('MONGO_QUEUE_APPLY_OPERATION_FAIL','MONGO_QUEUE_APPLY_OPERATION_FAIL');
+
 define('MONGO_CREATE_SEARCH_DOC','MONGO_CREATE_SEARCH_DOC');
 define('MONGO_CONNECTION_ERROR','MONGO_CONNECTION_ERROR');
 
+define('STAT_TYPE_COUNT', 'count');
```
robotrobot: I would rename this to SUBJECT_COUNT

rsinger:
New stats added by the branch:
- `MONGO_VIEW_CACHE_MISS`: an increment of requests for a View that needed to be regenerated from CBD
- `MONGO_TABLE_ROWS_DISTINCT`: an increment of Tables->distinct() calls
- `MONGO_QUEUE_DISCOVER_JOB.subjects_count`: keeps track of how many subjects are identified in a single DiscoverImpactedSubjects job
- `MONGO_QUEUE_DISCOVER_SUBJECT`: an increment of how many subjects total are impacted, also a timer
- `MONGO_QUEUE_DISCOVER_FAIL`: an increment of failed discover jobs
- `MONGO_QUEUE_APPLY_OPERATION.[operation_type]`: an increment of specific apply operation job for each subject (e.g. generate_table_rows, generate_views, search), also a timer
- `MONGO_QUEUE_APPLY_OPERATION_JOB.subject_count`: keeps track of how many subjects are sent to a single ApplyOperation job
- `MONGO_QUEUE_APPLY_OPERATION_FAIL`: an increment of failed apply jobs

PR #95, opened 2015-09-08
Composites should listen to read preference.

malcyL:
This change is to fix a problem where reading from Views is not listening to the read preference specified in the Driver.

After talking to @robotrobot, I have made the corresponding change for all composites - Views, Tables and SearchDocuments.

### DIFF
--- a/src/mongo/Driver.class.php
+++ b/src/mongo/Driver.class.php
@@ -119,6 +119,10 @@ class Driver extends DriverBase implements \Tripod\IDriver
 
         // is a custom stat tracker passed in?
         if ($opts['stat']!=null) $this->stat = $opts['stat'];
+
+        // Set the read preference if passed in
+        if ($opts['readPreference']) $this->readPreference = $opts['readPreference'];
+
     }
 
     /**
@@ -546,7 +550,8 @@ class Driver extends DriverBase implements \Tripod\IDriver
                 $this->storeName,
                 $this->collection,
                 $this->defaultContext,
-                $this->stat
+                $this->stat,
+                $this->readPreference
             );
         }
         return $this->tripod_views;
@@ -563,7 +568,8 @@ class Driver extends DriverBase implements \Tripod\IDriver
                 $this->storeName,
                 $this->collection,
                 $this->defaultContext,
-                $this->stat
+                $this->stat,
+                $this->readPreference
             );
         }
         return $this->tripod_tables;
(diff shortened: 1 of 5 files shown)


src/mongo/Driver.class.php
```diff
@@ -180,11 +180,12 @@ public function getViewForResources(Array $resources, $viewType)
     /**
      * @param array $filter
      * @param string $viewType
+     * @param string $readPreference
      * @return MongoGraph
      */
-    public function getViews(Array $filter, $viewType)
+    public function getViews(Array $filter, $viewType, $readPreference = \MongoClient::RP_PRIMARY_PREFERRED)
```
rsinger: I think this should default to `null` and then in the method:
```php 
if(is_null($readPreference))
{
    $readPreference = $this->readPreference;
}
```
I think it's counterintuitive if the developer has already set the default read preference when instantiating Tripod that the delegates wouldn't also respect that.

src/mongo/delegates/Views.class.php
```diff
@@ -120,9 +120,10 @@ public function getSpecification($storeName, $viewSpecId)
      * Return all views, restricted by $filter conditions, for given $viewType
      * @param array $filter - an array, keyed by predicate, to filter by
      * @param $viewType
+     * @param string $readPreference
      * @return \Tripod\Mongo\MongoGraph
      */
-    public function getViews(Array $filter,$viewType)
+    public function getViews(Array $filter, $viewType, $readPreference = \MongoClient::RP_PRIMARY_PREFERRED)
```
rsinger: Same comment as `Driver->getViews()`

src/mongo/Driver.class.php
```diff
@@ -184,7 +188,7 @@ public function getViewForResources(Array $resources, $viewType)
      */
     public function getViews(Array $filter, $viewType)
     {
-        return $this->getTripodViews()->getViews($filter,$viewType);
+        return $this->getTripodViews($this->readPreference)->getViews($filter,$viewType);
```
robotrobot: Why not have getTripodViews() access the read preference itself rather than rely on it being passed? Doing it this way is bug prone - and you have a bug. See other uses of getTripodViews() on lines 181 and 171 that don't pass the member - they won't use the read preference.

robotrobot: Actually looking at this I think this instance is at fault - the param has been left in and is not defined in the function - so I think removing the param here is the answer.

PR #97, opened 2015-09-29
90%: Fix queue stats

rsinger:
This PR brings in 2 changes:
1. it fixes a bug where queue stats weren't actually getting sent: `JobBase->getStat()` was using `isset($this->statConfig)` to determine if there was already a config set, but since it defaults to an array, it also needed to check if it was empty
2. It rips out the faulty understanding of the stats prefix and hard codes both "by store" (database) and aggregate stats with each stats send

### DIFF
--- a/src/classes/StatsD.class.php
+++ b/src/classes/StatsD.class.php
@@ -26,7 +26,7 @@ class StatsD implements ITripodStat
     {
         $this->host = $host;
         $this->port = $port;
-        $this->prefix = $prefix;
+        $this->setPrefix($prefix);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/classes/StatsD.class.php
```diff
@@ -192,4 +191,67 @@ public function setHost($host)
         $this->host = $host;
     }
 
+    /**
+     * @param string $operation
```
rgubby: Took a bit of jumping around function to function to work out what each one of these was doing.

It might be of value to add a desc here.

src/classes/StatsD.class.php
```diff
@@ -192,4 +191,67 @@ public function setHost($host)
...
+     */
+    public function getPivotValue()
+    {
+        return $this->pivotValue;
+    }
+
+    /**
+     * @param string $pivotValue
+     */
+    public function setPivotValue($pivotValue)
```
rgubby: Can't see where you could, but is it even possible to have a pivotValue of "."? Would that cause stats issues and do we need any checks on what this value can be?

rgubby:
:+1: looks good to me

src/classes/StatsD.class.php
rsinger: Hmm, interesting question.  As long as `.` isn't the leading or trailing character, I *think* it should be fine.  But some validation is warranted, will do.

PR #100, opened 2016-03-11
Add filtering and the inclusion of rdf sequences into views.

malcyL:
This pull request adds two new features to views.

1) It adds the ability to filter a view.

By filtering data included in the view, the view can be optimised to only include the required data. For example, the following view from the test data joins resources to other resources on the predicate dct:isVersionOf - but only if the resource is of type bibo:Chapter:

<pre>
                {
                    "_id": "v_resource_filter1",
                    "_version": "0.1",
                    "from": "CBD_testing",
                    "type": "acorn:Resource",
                    "include": ["rdf:type", "searchterms:topic"],
                    "joins": {
                        "dct:isVersionOf": {
                            "filter" : {
                                "rdf:type" : {"u" : "bibo:Chapter"}
                            },
                            "include": ["dct:subject", "rdf:type"]
                        }
                    }
                },

</pre>

While the view itself will only contain the filtered data, importantly the impact index is unchanged and still contains the id of all filtered subjects. This means that if a filtered subject is updated, the view will still be regenerated and the update taken into account in the filter. In the example above, a resource filtered out because it was a book, would appear in the view if it's type was changed to chapter.

2) It adds the ability to include an rdf sequence.

Currently, views can include individual properties, but it's not possible to include an rdf sequence because of the unknown number of triples. This pull requests adds a keyword which can be used in the include list to include an entire rdf sequence.

<pre>
                {
                    "_id": "v_resource_rdfsequence",
                    "_version": "0.1",
                    "from": "CBD_testing",
                    "type": "acorn:Resource",
                    "include": ["rdfSeq", "rdf:type"],
                    "joins": {
                        "dct:isVersionOf": {
                            "include": ["dct:subject", "rdf:type"]
                        }
                    }
                }
</pre>

By including "rdfSeq" the entire rdf sequence will be included in the view.


rsinger:
Would a viewspec with:
```
                            "filter" : {
                                "rdf:type.u" : "bibo:Chapter"
                            }
```

result in the same outcome?

rsinger:
The docs needs updating for this so it's clear what valid viewspecs can contain

rsinger:
My hunch is that it will automagically do the same thing.

rsinger:
Oh, actually, I guess not, looking at the filter code.

rsinger:
yeah, I was wondering about that, as well.

malcyL:
_seq_ works for me. I'll change it to that unless anyone suggest anything better?

rsinger:
I sort of feel that the filters shouldn't be limited 'predicate/object' only. I'm not suggesting they need to support *more* than predicate/object combination right now, but we should be designing it so you're explicitly saying you're filtering on that.  I could see us filtering on value >, <, etc. in the future.

src/mongo/delegates/Views.class.php
```diff
@@ -642,6 +672,39 @@ protected function extractProperties($source,$viewSpec,$from)
                 {
                     $obj[$p] = $source[$p];
                 }
+                if($p==='_seq_')
```
rsinger: `($p === '_seq_')` for style consistency.

robotrobot: Constant?

robotrobot: Also is `_link_` def the syntax we use elsehwere? Don't take my word for it.

rsinger:
Seems a little incompatible!  How would you know when one is a predicate and another is a count or something?

src/mongo/delegates/Views.class.php
```diff
@@ -599,8 +600,37 @@ protected function doJoins($source, $joins, &$dest, $from, $contextAlias, $build
...
                         if (isset($ruleset['joins'])) $this->expandSequence($ruleset['joins'],$linkMatch);
-
-                        $dest[_GRAPHS][] = $this->extractProperties($linkMatch,$ruleset,$from);
+                        if (isset($ruleset['filter']))
+                        {
+                            foreach($ruleset['filter'] as $filterPredicate => $filter)
+                            {
+                                foreach($filter as $filterType => $filterMatch)
+                                {
+                                    if (isset($linkMatch[$filterPredicate]))
```
rsinger: I think we might need to check for either the alias or the fquri here.  The data could potentially be inconsistent.

malcyL: The Views class has a labeller - adding this check isn't an issue - but...

How can the data be inconsistent? The $linkMatch will always contain the alias won't it? Would things work if it was the fquri? It seems to me it would have to be the alias.

And the $filterPredicate is something we're in control of when we specify the view - again which is going to be the alias.

Adding the check in is not a big thing, but how can the data be inconsistent like that?

rsinger: Aliases can be added at any time.  Until they're added, a FQ uri would be stored.  You're right about the link match, though.  I could have sworn there was an example where we look up either the alias or the fquri, but I can't find it, so I might be making the whole thing up.

malcyL: We're agreeing that $linkMatch is okay then. 

The $filterPredicate on this line is coming from the view. So what your saying is that someone could have specified the fquri for the filter predicate in the view? Something like:

```javascript
{
    "stores": {
        "some_store": {
            "pods" : {},
            "view_specifications": [
                {
                    "_id": "v_spec_1",
                    "from": "CBD_something",
                    "type": "owl:Thing"
                    "joins": {
                        "dct:isVersionOf": {
                            "filter" : {
                                "http://xmlns.com/foaf/0.1/name" : {"l" : "John"}
                            },
                            "include": ["dct:subject", "rdf:type"]
                        }
                    }
                }
            ]
        }
    }
}
```

i.e. the filter is using "http://xmlns.com/foaf/0.1/name" not "foaf:name".

Again - this is easy to do - we have a labeller in the view - but is this really supported for all the other parts of the view? Do we really want to support that here? Or am I misunderstanding what your saying?

malcyL:
I'm looking at how we could resolve @rsinger's comment above re filtering on more than just predicate/object in the future and ensuring the syntax we add here isn't incompatible with that.

The issue is 

```javascript
"filter" : {
  "rdf:type" : {"u" : "bibo:Chapter"}
}
```

vs

```javascript
"filter" : {
  "item:count" : {"l" : { $gte:30 } }
}
```

Is it worth noting that the filter is *not* used in a mongo query. We have to maintain all filtered items in the impact index so we can not just add the filter into the mongo query. We have to bring all items that could potentially be in the view and filter in memory. So any future filters we add will have to implement  this feature - not delegate it to mongo.

Do we need to specify the type of the filter:

```javascript
"filters" : [
  {
    "rdf:type" : "tripod:PredicateObjectFilter",
    "filter" : {
      "rdf:type" : {"u" : "bibo:Chapter"}
    }
  },
  {
    "rdf:type" : "tripod:CountFilter",
    "filter" : {
      "item:count" : {"l" : { $gte:30 } }
    }
  }
]
```

Thoughts @robotrobot @rsinger?

malcyL:
Talking to @robotrobot, I think I (and possibly @rsinger?) have misunderstood @robotrobot's reply above:

> @rsinger what we have doesn't seem incompatible with that.
> 
> "rdf:type": {u:"somepred"} could become "item:count": {l:{$gte:30}} later.
> 
 
@rsinger replied:

> Seems a little incompatible! How would you know when one is a predicate and another is a count or something?

The "item:count" in the example is intended as a predicate. So it would always be some predicate matching against some filter value. But whatever that filter value is - it's not incompatible with the syntax we have.

So I think - we're happy that what we have doesn't cause a problem for implementing different filters in the future. Are you okay with this @rsinger ?

src/mongo/delegates/Views.class.php
malcyL: Talking to @robotrobot is the problem not on this line, but a little below where we check the $linkMatchValue.

My plan is to leave this line unchanged - I don't believe we need to check for a fquri for either the $linkMatch or the $filterPredicate here.

However, I will add this to the $linkMatchValue lower down.

rsinger: What I meant by my comment was that you were right and I was wrong.  We _can_ have a mix of aliases and fquris in the database, but it's not going to make any difference here, because we'd only be matching whatever is specified in the spec.

rsinger:
I see.  Ok.

malcyL:
Verbal :+1: from @robotrobot - I'm going to merge this.

PR #102, opened 2016-05-31
20% Outdated spec regeneration

junglebarry:
This PR is still work-in-progress, but at a point where feedback is sought.
# Aims
- Adds a `revision` to composite specifications, a monotonically-increasing integer
- When a composite document is being saved, include the specification revision in the document
- Provide a query to find composite documents that are generated with older (outdated) specifications
- Regenerates outdated composite documents

The current code deals only with views, but should generalise to other composites.  Some of the mechanisms are already generic enough to support this, but where they are not, I've attempted to provide more abstract mechanisms.  There are some that need reworking.

The bulk of the code is in a new job, called `DiscoverOutdatedComposites`.  This is designed to run on a CRON schedule, and at each `perform` step, will iterate across all composites, find a list of documents that need regeneration, and trigger their regeneration.  The current code does this all inside `perform` for a single job, but the ultimate intention is to split out the regeneration from the discovery.

No effort has been made to include logging or profiling at this point.  No effort has been made to trigger jobs.  Testing is unit-test only.
## Representing Revisions

The code currently uses independent revision numbers for every composite spec.  This is simple, and makes comparisons easy, so documents need regeneration when:
- For a single document, its `revision` is less than the revision of its specification;
- For a query, we can just use mongodb's `$lt` operator to extract all matches.

It's also easy for distribution, since outdated replicas will naturally stop contributing (eventually).

Despite this simplicity, this revisioning scheme has issues:
- It's a lot of manual work to manage spec updates;
- It is hard to automate, because we need to retain a history of changes, making build-time revision update difficult (we need to store history somewhere external - a file of `<specification, revision, hash>` triples, perhaps?)
### Alternative: single revision number, plus per-specification hashes

The original suggestion from Chris was to use a hash, since it picks up semantic changes only, and is easy to compute.  Without some notion of order, however, it's possible to get into "tennis matches", where two replicas disagree on the correct hash, and attempt to undo each other's work.  By adding in a single global revision (again monotonically-increasing), we can provide the order needed to resolve this problem.

Using a hash makes comparisons more complex, so documents need regeneration when:
- For a single document, its `revision` is less than the global revision **and** its hash is different to the hash of its specification;
- For a query, we need `{ _revision: { $lt: <GLOBAL REVISION> }, _hash: { $ne: <SPEC. HASH> }}`

Note that the revision is checked first in both cases, so takes precedence over the hash.

The hash-based approach has a couple of nice properties:
- We can trivially automate incrementing the global revision (since we don't need to compare with previous versions - the hash will do that)
- The hash only captures semantic changes
### Composite classes are tied to podNames

In order to construct a `Views` instance, I need a collection associated with a `podName`, which is broadly unused.  Within each composite, the collection/podName are rarely used, but they are enforced by the hierarchy (`Views <-- CompositeBase <-- DriverBase`).  It cannot be simply removed.

I'm not sure that Views/Tables/Search is conceptually tied to a pod.  Could do with some discussion here to figure out why it is tied thus at present.
### Regeneration is not uniformly available

We really need to create a generic mechanism for regenerating a document from its spec and root CBD, defined at the level of `IComposite`.  I've added another interface (`ICompositeRegen`) for now, so it need not be immediately applied to all composite types.
## TODO
- [ ] Remove comments that are present only for discussion in the PR;
- [ ] Settle the revision/hash debate
- [ ] Split the discovery and regeneration jobs on to separate queues
- [x] Abstract the regeneration code into (a subset of) `IComposite` implementors.
- [ ] Fetch all types of composites during the discovery step
- [ ] Add logging back in - especially for counts of outdated documents
- [ ] Add profiling

### DIFF
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -2064,6 +2064,48 @@ class Config
     }
 
     /**
+    * Creates a specification revision subdocument for a composite specification.
+    *
+    * @param \MongoCollection $specification - the composite specification
+    * @return array|null - the subdocument to add to the composite
+    **/
+    public static function getSpecRevisionSubDoc($specification) {
+        if(isset($specification) && isset($specification[_ID_KEY]) && isset($specification[_REVISION])) {
+            return array(
+                _SPEC_KEY => array(
+                    _SPEC_TYPE=> $specification[_ID_KEY],
+                    _SPEC_REVISION => $specification[_REVISION]
+                )
+            );
+        } else {
+            return null;
+        }
+    }
+
+    /**
+    * Ensures that specification revision is indexed for a composite collection.
+    *
+    * Note that spec revision is not required for all composite documents, so
+    * a sparse index is used to avoid indexing unrevisioned documents.
+    *
+    * @param \MongoCollection $compositeCollection - the composite collection
+    **/
+    public static function ensureIndexForSpecRevision($compositeCollection) {
+        if(isset($compositeCollection)) {
+            $compositeCollection->ensureIndex(
+                array(
+                    _SPEC_KEY.'.'._SPEC_TYPE => 1,
+                    _SPEC_KEY.'.'._SPEC_REVISION => 1
+                ),
+                array(
+                    'background' => 1,
+                    'sparse' => 1
+                )
+            );
+        }
+    }
+
+    /**
      * @return string
      */
     public static function getDiscoverQueueName()
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1795,6 +1795,18 @@ class Config
 
     /**
      * @param string $storeName
+     * @param Array $spec
+     * @param string $readPreference
+     * @throws \Tripod\Exceptions\ConfigException
+     * @return \MongoCollection
+     */
+    public function getFromCollectionForSpec($storeName, $spec, $readPreference = \MongoClient::RP_PRIMARY_PREFERRED)
+    {
+        return $this->getCollectionForCBD($storeName, $spec['from'], $readPreference);
+    }
+
+    /**
+     * @param string $storeName
      * @param string $viewId
      * @param string $readPreference
      * @throws \Tripod\Exceptions\ConfigException
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1806,6 +1806,32 @@ class Config
     }
 
     /**
+     * Gets the collection containing documents for a given composite type.
+     * @param string $compositeType
+     * @param string $storeName
+     * @param string $viewId
+     * @param string $readPreference
+     * @throws \Tripod\Exceptions\ConfigException
+     * @return \MongoCollection
+     */
+    public function getCollectionForCompositeType($compositeType, $storeName, $compositeId, $readPreference = \MongoClient::RP_PRIMARY_PREFERRED)
+    {
+        switch ($compositeType) {
+            case COMPOSITE_TYPE_VIEWS:
+                return $this->getCollectionForView($storeName, $compositeId, $readPreference);
+
+            case COMPOSITE_TYPE_TABLES:
+                return $this->getCollectionForTable($storeName, $compositeId, $readPreference);
+
+            case COMPOSITE_TYPE_SEARCH:
+                return $this->getCollectionForSearchDocument($storeName, $compositeId, $readPreference);
+
+            default:
+                throw new \Tripod\Exceptions\ConfigException("Undefined composite type '$compositeType' requested");
+        }
+    }
+
+    /**
      * @param string $storeName
      * @param string $viewId
      * @param string $readPreference
--- a/src/mongo/ICompositeRegen.php
+++ b/src/mongo/ICompositeRegen.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Tripod\Mongo\Composites;
+
+/**
+ * Interface ICompositeRegen
+ * @package Tripod\Mongo\Composites
+ */
+interface ICompositeRegen
+{
+    /**
+     * Regenerate a specific composite document from its spec and root CBD
+     * @param $specification
+     * @param $compositeCollection
+     * @param $rootCbd
+     * @return void
+     */
+    public function regenerateOne($specification, $compositeCollection, $rootCbd);
+}
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -406,7 +406,7 @@ class Views extends CompositeBase
             $t = new \Tripod\Timer();
             $t->start();
 
-            $from = $this->getFromCollectionForViewSpec($viewSpec);
+            $from = $viewSpec['from'];
             $collection = $this->config->getCollectionForView($this->storeName, $viewId);
 
             $this->ensureIndices($viewSpec, $collection);
@@ -509,15 +509,7 @@ class Views extends CompositeBase
         );
 
         // spec type/revision index lets us find outdated views
-        $collection->ensureIndex(
-            array(
-                _SPEC_KEY.'.'._SPEC_TYPE => 1,
-                _SPEC_KEY.'.'._SPEC_REVISION => 1
-            ),
-            array(
-                'background'=>1
-            )
-        );
+        Config::ensureIndexForSpecRevision($collection);
 
         // ensure any custom view indexes
         if (isset($viewSpec['ensureIndexes']))
@@ -534,6 +526,18 @@ class Views extends CompositeBase
         }
     }
 
+    /**
+    * Constructs a view document to save from the relevant root CBD document, and saves it.
+    *
+    * TODO: we must only save if $viewSpec[_revision] >= current saved view document spec.revision
+    * This will ensure that view specification revisions monotonically increase.
+    *
+    * @param array $viewSpec
+    * @param \MongoCollection $collection - the views composite collection
+    * @param array $doc - the root CBD from which this view is generated
+    * @param string $from - the name of the CBD collection from which $doc is drawn
+    * @param array $contextAlias - the context for this view
+    **/
     public function saveGeneratedView($viewSpec, $collection, $doc, $from, $contextAlias) {
         // set up ID
         $generatedView = array(
@@ -541,12 +545,15 @@ class Views extends CompositeBase
                 _ID_RESOURCE=>$doc["_id"][_ID_RESOURCE],
                 _ID_CONTEXT=>$doc["_id"][_ID_CONTEXT],
                 _ID_TYPE=>$viewSpec['_id']
-            ),
-            _SPEC_KEY => array(
-                _SPEC_TYPE=> $viewSpec['_id'],
-                _SPEC_REVISION => $viewSpec[_REVISION]
             )
         );
+
+        // if specification is revisioned, add revision information to what we will save
+        $specRevision = Config::getSpecRevisionSubDoc($viewSpec);
+        if(isset($specRevision)) {
+            $generatedView = array_merge($generatedView, $specRevision);
+        }
+
         $value = array(); // everything must go in the value object todo: this is a hang over from map reduce days, engineer out once we have stability on new PHP method for M/R
 
         $value[_GRAPHS] = array();
@@ -859,24 +866,6 @@ class Views extends CompositeBase
         }
 
         return $obj;
-    }
-
-    /**
-     * @param string $viewSpec
-     * @return string
-     */
-    private function getFromCollectionForViewSpec($viewSpec)
-    {
-        $from = null;
-        if (isset($viewSpec["from"]))
-        {
-            $from = $viewSpec["from"];
-        }
-        else
-        {
-            $from = $this->podName;
-        }
-        return $from;
     }
 
     /**
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -12,7 +12,7 @@ use Tripod\Mongo\Labeller;
  * Class Views
  * @package Tripod\Mongo\Composites
  */
-class Views extends CompositeBase
+class Views extends CompositeBase implements ICompositeRegen
 {
 
     /**
@@ -877,4 +877,15 @@ class Views extends CompositeBase
         return $this->getConfigInstance()->getCollectionForView($this->storeName, $viewSpecId);
     }
 
+    
+    /**
+     * Regenerate a specific composite document from its spec and root CBD
+     * @param $specification
+     * @param $compositeCollection
+     * @param $rootCbd
+     * @return void
+     */
+    public function regenerateOne($specification, $compositeCollection, $rootCbd) {
+        $this->saveGeneratedView($specification, $compositeCollection, $rootCbd, $specification['from'], $this->defaultContext);
+    }
 }
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -536,7 +536,7 @@ class Views extends CompositeBase
     * @param \MongoCollection $collection - the views composite collection
     * @param array $doc - the root CBD from which this view is generated
     * @param string $from - the name of the CBD collection from which $doc is drawn
-    * @param array $contextAlias - the context for this view
+    * @param string $contextAlias - the context for this view
     **/
     public function saveGeneratedView($viewSpec, $collection, $doc, $from, $contextAlias) {
         // set up ID
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -19,7 +19,7 @@ class Views extends CompositeBase
      * Construct accepts actual objects rather than strings as this class is a delegate of
      * Tripod and should inherit connections set up there
      * @param string $storeName
-     * @param \MongoCollection $collection
+     * @param \MongoCollection $collection // TODO: should this really be a constructor arg?
      * @param $defaultContext
      * @param null $stat
      * @param string $readPreference
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -169,6 +169,7 @@ class Views extends CompositeBase
         {
             $this->getStat()->increment(MONGO_VIEW_CACHE_MISS.".$viewType");
             $viewSpec = Config::getInstance()->getViewSpecification($this->storeName, $viewType);
+            // TODO: remove this line once discussion closed.
             if($viewSpec == null)
             {
                 return new \Tripod\Mongo\MongoGraph();
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -175,7 +175,9 @@ class Views extends CompositeBase
                 return new \Tripod\Mongo\MongoGraph();
             }
 
-            $fromCollection = $this->getFromCollectionForViewSpec($viewSpec);
+
+            $fromCollection = $viewSpec['from'];
+
 
             $doc = $this->config->getCollectionForCBD($this->storeName, $fromCollection)
                 ->findOne(array( "_id" => array("r"=>$resourceAlias,"c"=>$contextAlias)));
@@ -203,8 +205,7 @@ class Views extends CompositeBase
     private function getCBDCollectionForViewType($viewType) {
         $config = $this->getConfigInstance();
         $viewSpec = $config->getViewSpecification($this->storeName, $viewType);
-        $fromCollection = $this->getFromCollectionForViewSpec($viewSpec);
-        return $config->getCollectionForCBD($this->storeName, $fromCollection);
+        return $config->getFromCollectionForSpec($this->storeName, $viewSpec);
     }
 
     /**
@@ -408,53 +409,11 @@ class Views extends CompositeBase
             $from = $this->getFromCollectionForViewSpec($viewSpec);
             $collection = $this->config->getCollectionForView($this->storeName, $viewId);
 
+            $this->ensureIndices($viewSpec, $collection);
+
             if (!isset($viewSpec['joins']))
             {
                 throw new \Tripod\Exceptions\ViewException('Could not find any joins in view specification - usecase better served with select()');
-            }
-
-            // ensure that the ID field, view type, and the impactIndex indexes are correctly set up
-            $collection->ensureIndex(
-                array(
-                    '_id.r'=>1,
-                    '_id.c'=>1,
-                    '_id.type'=>1
-                ),
-                array(
-                    'background'=>1
-                )
-            );
-
-            $collection->ensureIndex(
-                array(
-                    '_id.type'=>1
-                ),
-                array(
-                    'background'=>1
-                )
-            );
-
-            $collection->ensureIndex(
-                array(
-                    'value.'._IMPACT_INDEX=>1
-                ),
-                array(
-                    'background'=>1
-                )
-            );
-
-            // ensure any custom view indexes
-            if (isset($viewSpec['ensureIndexes']))
-            {
-                foreach ($viewSpec['ensureIndexes'] as $ensureIndex)
-                {
-                    $collection->ensureIndex(
-                        $ensureIndex,
-                        array(
-                            'background'=>1
-                        )
-                    );
-                }
             }
 
             $types = array(); // this is used to filter the CBD table to speed up the view creation
@@ -503,31 +462,7 @@ class Views extends CompositeBase
                 }
                 else
                 {
-                    // set up ID
-                    $generatedView = array("_id"=>array(_ID_RESOURCE=>$doc["_id"][_ID_RESOURCE],_ID_CONTEXT=>$doc["_id"][_ID_CONTEXT],_ID_TYPE=>$viewSpec['_id']));
-                    $value = array(); // everything must go in the value object todo: this is a hang over from map reduce days, engineer out once we have stability on new PHP method for M/R
-
-                    $value[_GRAPHS] = array();
-
-                    $buildImpactIndex=true;
-                    if (isset($viewSpec['ttl']))
-                    {
-                        $buildImpactIndex=false;
-                        $value[_EXPIRES] = new \MongoDate($this->getExpirySecFromNow($viewSpec['ttl']));
-                    }
-                    else
-                    {
-                        $value[_IMPACT_INDEX] = array($doc['_id']);
-                    }
-
-                    $this->doJoins($doc,$viewSpec['joins'],$value,$from,$contextAlias,$buildImpactIndex);
-
-                    // add top level properties
-                    $value[_GRAPHS][] = $this->extractProperties($doc,$viewSpec,$from);
-
-                    $generatedView['value'] = $value;
-
-                    $collection->save($generatedView);
+                    $this->saveGeneratedView($viewSpec, $collection, $doc, $from, $contextAlias);
                 }
             }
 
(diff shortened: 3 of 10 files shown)


src/mongo/delegates/Views.class.php
```diff
@@ -169,6 +169,7 @@ public function getViewForResource($resource,$viewType,$context=null)
         {
             $this->getStat()->increment(MONGO_VIEW_CACHE_MISS.".$viewType");
             $viewSpec = Config::getInstance()->getViewSpecification($this->storeName, $viewType);
+            // TODO: remove this line once discussion closed.
```
junglebarry: This check exists in `getViewForResource`, but the equivalent check does not exist in `getViewForResources`.  It seems to me that whatever conditions hold in the formert case must also hold in the latter, so should both implement the check?  I noticed this while attempting a small refactor of the common components of each method.

src/mongo/delegates/Views.class.php
```diff
@@ -217,25 +229,22 @@ public function getViewForResources(Array $resources,$viewType,$context=null)
         $missingSubjects = array_diff($resources,$returnedSubjects);
         if (!empty($missingSubjects))
         {
+            $cbdCollection = $this->getCBDCollectionForViewType($viewType);
+            $contextAlias = $this->getContextAlias($context);
+
             $regrabResources = array();
             foreach($missingSubjects as $missingSubject)
             {
-                $viewSpec = $this->getConfigInstance()->getViewSpecification($this->storeName, $viewType);
```
junglebarry: I doubt this code is expensive, but it is happening on every iteration of the loop.  I pulled it outside the loop for efficiency (and because I wanted to rationalise the code for reuse in other view-regeneration methods).

src/mongo/delegates/Views.class.php
junglebarry: Sorry, the check is the block immediately below.

src/mongo/delegates/Views.class.php
```diff
@@ -217,25 +229,22 @@ public function getViewForResources(Array $resources,$viewType,$context=null)
...
-                $fromCollection = $this->getFromCollectionForViewSpec($viewSpec);
-
                 $missingSubjectAlias = $this->labeller->uri_to_alias($missingSubject);
-                $doc = $this->getConfigInstance()->getCollectionForCBD($this->storeName, $fromCollection)
+                $doc = $cbdCollection
                     ->findOne(array( "_id" => array("r"=>$missingSubjectAlias,"c"=>$contextAlias)));
 
-                if($doc == NULL)
+                if(!is_null($doc))
```
junglebarry: Do we have a reason to prefer `== NULL` over the predicate?

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+        $storeName = $this->args[self::STORE_NAME_KEY];
+        $cursorLimit = $this->args[self::CURSOR_LIMIT];
+
+        // set the config to what is received
+        \Tripod\Mongo\Config::setConfig($config);
+
+        // closure around $cursorLimit in $this->getRegenTaskForMetadata
+        $getRegenTaskForMetadata = function($metadatum) use ($cursorLimit) { 
+            return $this->getRegenTaskForMetadata($metadatum, $cursorLimit); 
+        };
```
junglebarry: Perhaps split the one job per spec, for fine-grained control.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+            $cbdCollection = $metadatum->cbdCollection;
+            $compositeRegenFunction = $metadatum->compositeRegenFunction;
+
+            $cbdIdFields = array(_ID_KEY.'.'._ID_RESOURCE => 1, _ID_KEY.'.'._ID_CONTEXT => 1);
+
+            // find one cursor page at a time for this composite
+            $outdatedComposites = 
+                $compositeCollection
+                    ->find($filterOutdated, $cbdIdFields)
+                    ->limit($cursorLimit);
```
junglebarry: Can we guarantee that we don't get starvation on rare data?

We should definitely track and log the number of records that are out-of-date for a given composite, so we can observe behaviour.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+            if(!$outdatedComposites->hasNext()) {
+                // no outdated composite documents found, so no work to be done
+                return null;
+            } else {
+                // find IDs root CBDs required to regenerate the composite
+                $outdatedCbdIds = 
+                    array_map(array($this, 'composite2cbdId'), iterator_to_array($outdatedComposites, false));
+
+                // fetch the CBDs themselves
+                $filterCbdsById = array('$or' => $outdatedCbdIds);
```
junglebarry: Is it possible to starve certain composite types, if a frequently-updated version is always before it in the `$or`?

src/mongo/delegates/Views.class.php
```diff
@@ -399,55 +409,13 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
...
             $collection = $this->config->getCollectionForView($this->storeName, $viewId);
 
+            $this->ensureIndices($viewSpec, $collection);
+
             if (!isset($viewSpec['joins']))
             {
                 throw new \Tripod\Exceptions\ViewException('Could not find any joins in view specification - usecase better served with select()');
             }
 
-            // ensure that the ID field, view type, and the impactIndex indexes are correctly set up
```
junglebarry: This whole block is moved, verbatim, with a single addition marked.

src/mongo/delegates/Views.class.php
```diff
@@ -494,31 +462,7 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
                 }
                 else
                 {
-                    // set up ID
```
junglebarry: This block is moved to a new method, for later reuse.

src/mongo/delegates/Views.class.php
```diff
@@ -532,6 +476,102 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
...
+        $collection->ensureIndex(
+            array(
+                'value.'._IMPACT_INDEX=>1
+            ),
+            array(
+                'background'=>1
+            )
+        );
+
+        // spec type/revision index lets us find outdated views
```
junglebarry: This has been added, to allow us to search composite documents by spec revision.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+                        $specification, 
+                        $compositeCollection, 
+                        iterator_to_array($cbdDocs, false)
+                    );
+                }
+            }
+        }
+    }
+
+    // TODO: this should really live in IComposite implementations, but cannot yet.
```
junglebarry: This entire function should really live in an `IComposite` implementation, like `Views.class.php`.  However, I've put it here for now so we can discuss my concerns.

It's currently difficult to move it there because their constructors are tied to collections that seem independent of all views, and contexts that seem like a function-level specification.

src/mongo/delegates/Views.class.php
```diff
@@ -822,24 +869,6 @@ protected function extractProperties($source,$viewSpec,$from)
     }
 
     /**
-     * @param string $viewSpec
-     * @return string
-     */
-    private function getFromCollectionForViewSpec($viewSpec)
```
junglebarry: The validation says that `"from"` is a mandatory field on view specifications, so there is no eventuality when we should fall back on `$this->podName`.  Removing, in favour of fetching the `'from'` collection name from the spec directly.

src/mongo/delegates/Views.class.php
```diff
@@ -19,7 +19,7 @@ class Views extends CompositeBase
      * Construct accepts actual objects rather than strings as this class is a delegate of
      * Tripod and should inherit connections set up there
      * @param string $storeName
-     * @param \MongoCollection $collection
+     * @param \MongoCollection $collection // TODO: should this really be a constructor arg?
```
junglebarry: I can't fathom why this is a constructor argument.

`Views` is a class that is used to handle all views, and as such is not tied to a single collection.  Indeed, many of the operations operate directly on the `"views"` collection, rather than a CBD collection.  However, in order to construct a `Views` instance, you need to pass a CBD collection.  This seems to be overly coupled with the use of `Views` from a tripod instance, but it hinders other use cases.

The collection and associated podName are barely used in this class, and even then, only in `generateViews`, which is only invoked thus:

```php
public function update(\Tripod\Mongo\ImpactedSubject $subject)
    {
        $resource = $subject->getResourceId();
        $resourceUri    = $resource[_ID_RESOURCE];
        $context        = $resource[_ID_CONTEXT];

        $this->generateViews(array($resourceUri),$context);
    }
```

The `ImpactedSubject` has the podName, so it would be perfectly possible to fetch this collection and pass as an argument to `generateViews`.

How would y'all feel about me removing this constructor argument?

src/mongo/delegates/Views.class.php
```diff
@@ -396,58 +406,16 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
             $t = new \Tripod\Timer();
             $t->start();
 
-            $from = $this->getFromCollectionForViewSpec($viewSpec);
+            $from = $viewSpec['from'];
             $collection = $this->config->getCollectionForView($this->storeName, $viewId);
 
+            $this->ensureIndices($viewSpec, $collection);
```
junglebarry: This seems to be happening on every update.  Is that necessary?  Once configured, the indices should take care of themselves...

src/mongo/delegates/Views.class.php
```diff
@@ -532,6 +476,109 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
...
+
+        // ensure any custom view indexes
+        if (isset($viewSpec['ensureIndexes']))
+        {
+            foreach ($viewSpec['ensureIndexes'] as $ensureIndex)
+            {
+                $collection->ensureIndex(
+                    $ensureIndex,
+                    array(
+                        'background'=>1
```
junglebarry: Should these be sparse indexes?  If they index properties that are specific to a subset of documents, it seems wasteful to index everything.

src/mongo/delegates/Views.class.php
```diff
@@ -532,6 +476,109 @@ public function generateView($viewId,$resource=null,$context=null,$queueName=nul
...
+                _ID_CONTEXT=>$doc["_id"][_ID_CONTEXT],
+                _ID_TYPE=>$viewSpec['_id']
+            )
+        );
+
+        // if specification is revisioned, add revision information to what we will save
+        $specRevision = Config::getSpecRevisionSubDoc($viewSpec);
+        if(isset($specRevision)) {
+            $generatedView = array_merge($generatedView, $specRevision);
+        }
```
junglebarry: The above code adds in the revision identifier, if this spec is revisioned.

junglebarry: We should only save this document if its revision is the greater-than-or-equal-to the revision of the existing document in the database.  If we do not ensure this, we can overwrite our document with copies generated by an old revision of the spec.  We must prevent that.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+    public function makeViewRegenFunc($storeName) {
+        // TODO: remove $views from here. We need it because that class provides a method for
+        // regenerating indvidual views.  However, that should probably be abstracted somewhere,
+        // into a method capable of regenerating individual composites, by type.
+
+        // TODO: these fakes are only here to satisfy Views, which probably does not
+        // need either of these as constructor arguments (most methods in Views are agnostic)
+        $fakeCollection = \Tripod\Mongo\Config::getInstance()->getCollectionForCBD($storeName, 'CBD_testing');
+        $fakeDefaultContext = 'http://talisaspire.com/';
```
junglebarry: These are hard-coded to demonstrate issues with constructor calls, and must not make it into production.

* [ ] Remove hard-coded values for Views constructor

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,296 @@
...
+            return null;
+        } else {
+            $compositeCollection = $metadatum->compositeCollection;
+            $specification = $metadatum->specification;
+            $cbdCollection = $metadatum->cbdCollection;
+            $compositeRegenFunction = $metadatum->compositeRegenFunction;
+
+            $cbdIdFields = array(_ID_KEY.'.'._ID_RESOURCE => 1, _ID_KEY.'.'._ID_CONTEXT => 1);
+
+            // find one cursor page at a time for this composite
```
junglebarry: The following query finds one cursor of outdated documents for the specified metadata object.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,318 @@
...
+    * It would also be possible to run this query using a single global revision and a composite spec hash:
+    *
+    * { 
+    *   "_spec.type": <THIS TYPE>, 
+    *   "_spec.revision": { "$lt": <GLOBAL REVISION NUMBER> }, 
+    *   "_spec.hash": { "$ne": <HASH OF THE SPEC> }
+    * }
+    *
+    * That alternative would make managing the revision numbers much more straightforward, since we can simply
+    * Increment that global number on every build, and the hash will tell us if an update is required. Crucially,
```
junglebarry: The above method is doing most of the work for our database lookups, and deserves special scrutiny.

The method constructs a query component, which can be used to filter to outdated documents in the metadata's composite collection.

junglebarry: We should alter behaviour to better handle moving a specification from no revision to its first revision:

* Any spec that currently has a null revision has no outdated documents, so if the revision is null, no component;
* `$lt` will not cover nulls, so we need to add a match against unrevisioned composite documents.

src/mongo/delegates/Views.class.php
junglebarry: An upsert would be preferable to a compare-and-set, so that it's kept atomic.

src/mongo/Config.class.php
```diff
@@ -2026,6 +2064,48 @@ public function getTransactionLogDatabase($readPreference = \MongoClient::RP_PRI
     }
 
     /**
```
junglebarry: These functions need not live in `Config`; they might be better placed in some kind of composite utilities class.  They're portable, though.

rsinger:
> I'm not sure that Views/Tables/Search is conceptually tied to a pod.

I don't know that I follow this bit.  I understand that explicitly passing the pod is possibly redundant because you can theoretically get the 'starting' pod from the composite's config, but all specs start *somewhere*.  That is, your composite may contain a bunch of joins to various other pods, but they all start with, and revolve around, a specific graph document and that is always tied to an individual pod.  In the case of views, specifically, there's a decent chance that Tripod will have to regenerate the view (due to TTLs) on the fly - really the view is a pre-hydrated representation of a specific resource, so I don't really see how it would be fundamentally any different than how you would retrieve the raw resource graph.

What scenario do you picture that doesn't conform to that?

src/mongo/Config.class.php
rsinger: What's the ETA on TARL migrating to PHP 5.4+?  Perhaps it makes more sense to make 'composite-ness' a trait.

src/mongo/ICompositeRegen.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Tripod\Mongo\Composites;
+
+/**
+ * Interface ICompositeRegen
+ * @package Tripod\Mongo\Composites
+ */
+interface ICompositeRegen
```
rsinger: Why would this not just be part of IComposite?

src/mongo/Config.class.php
robotrobot: Not in this PR's lifetime :-(

robotrobot: Could these go in a CompositeBase class?

src/mongo/Config.class.php
```diff
@@ -2026,6 +2064,48 @@ public function getTransactionLogDatabase($readPreference = \MongoClient::RP_PRI
...
+    * Creates a specification revision subdocument for a composite specification.
+    *
+    * @param \MongoCollection $specification - the composite specification
+    * @return array|null - the subdocument to add to the composite
+    **/
+    public static function getSpecRevisionSubDoc($specification) {
+        if(isset($specification) && isset($specification[_ID_KEY]) && isset($specification[_REVISION])) {
+            return array(
+                _SPEC_KEY => array(
+                    _SPEC_TYPE=> $specification[_ID_KEY],
```
robotrobot: What is `_SPEC_TYPE`? Should it not be `_SPEC_ID` if it is read from `$specification[_ID_KEY]`

src/mongo/delegates/Views.class.php
robotrobot: +1

src/mongo/delegates/Views.class.php
robotrobot: Generally, I'd agree with this. Or in both cases throw an exception that you've requested a view type that doesn't exist.

src/mongo/delegates/Views.class.php
robotrobot: Yup agree, they should def be sparse as not all views will have them (they are type specific)

src/mongo/delegates/Views.class.php
rsinger: This has come up before.  The issue is that it's not entirely clear how you would ensure that the indexes have been created.  That is, what process would exist to make sure this happens and, if it hasn't, does it?

As it is, `ensureIndex` isn't terribly expensive, so the feeling is that it's less expensive to ensureIndexes rather than miss them.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,323 @@
...
+        // compile a list of all composite configuration metadata objects
+        $compositeMetadata = $this->getCompositeMetadata($config, $storeName);
+
+        // for each composite metadata instance, fetch regen tasks (where they exist), 
+        // which include specific CBDs to regenerate
+        $regenTasksIncludingNulls = array_map($getRegenTaskForMetadata, $compositeMetadata);
+        $regenTasks = array_filter($regenTasksIncludingNulls, 'isset');
+
+        // for collected regeneration tasks, run the update
+        // TODO: this will probably want to schedule updates, rather than running directly
```
robotrobot: In a different job?

src/mongo/delegates/Views.class.php
junglebarry: Ok.  I can see that there's not an "on deployment, run this" type of job where we could naturally hook that in, so I accept this approach is sensible.  Thanks for clarifying!

Would you object to me centralising the shared indices to reduce code, as I have done with `Config::ensureIndexForSpecRevision`?

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,323 @@
...
+     */
+    public function createJob(Array $data, $queueName=null)
+    {
+        if(!$queueName)
+        {
+            $queueName = Config::getDiscoverQueueName();
+        }
+        elseif(strpos($queueName, \Tripod\Mongo\Config::getDiscoverQueueName()) === false)
+        {
+            $queueName = \Tripod\Mongo\Config::getDiscoverQueueName() . '::' . $queueName;
```
robotrobot: this is namespacing with `getDiscoverQueueName()` if `$queueName` is not already doing so, right?

src/mongo/ICompositeRegen.php
junglebarry: Ultimately, I think it will (by the end of this PR).  Separate lets me do it only for Views until we're happy with the approach, then migrated it into `IComposite`.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
robotrobot: Can this be done using the stat in tripod? https://github.com/talis/tripod-php/blob/master/src/classes/StatsD.class.php

src/mongo/jobs/DiscoverOutdatedComposites.class.php
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    /**
+    * Converts a composite document into an ID for finding its root CBD.
+    *
+    * @param array $compositeDocument - the composite being examined
+    * @return array - the ID of $compositeDocument's root CBD.
+    **/
+    protected function composite2cbdId($compositeDocument) {
```
robotrobot: `2` => `To` and `cbd` => `CBD`

test/unit/mongo/DiscoverOutdatedCompositesTest.php
```diff
@@ -0,0 +1,360 @@
...
+    protected $tripodViews = null;
+
+    protected function setUp()
+    {
+        parent::setup();
+
+        $this->tripodTransactionLog = new \Tripod\Mongo\TransactionLog();
+        $this->tripodTransactionLog->purgeAllTransactions();
+
+        // Stub ouf 'addToElastic' search to prevent writes into Elastic Search happening by default.
```
robotrobot: This is an outdated copy-and-pasted comment, it references Elastic Search.

junglebarry:
> > I'm not sure that Views/Tables/Search is conceptually tied to a pod.
> 
> I don't know that I follow this bit. I understand that explicitly passing the pod is possibly redundant because you can theoretically get the 'starting' pod from the composite's config, but all specs start somewhere. That is, your composite may contain a bunch of joins to various other pods, but they all start with, and revolve around, a specific graph document and that is always tied to an individual pod. In the case of views, specifically, there's a decent chance that Tripod will have to regenerate the view (due to TTLs) on the fly - really the view is a pre-hydrated representation of a specific resource, so I don't really see how it would be fundamentally any different than how you would retrieve the raw resource graph.
> 
> What scenario do you picture that doesn't conform to that?

I agree that every composite has a root CBD, and that is defined by the pod.

The scenario I have is that I'm starting from *all* composite specifications, and each one could have a different pod at its root.  When you start here, you want to perform actions that are generic across all composites, and therefore you'd have to create many new instances of `Views` on the way through.

However, I suspect that if I split the jobs up by composite, I'll have a more natural fit for `Views`-tied-to-a-pod.  Let's hold fire on that until the discussion.

src/mongo/delegates/Views.class.php
rsinger: Yep.  This was a point in time method that has long-since passed.

test/unit/mongo/DiscoverOutdatedCompositesTest.php
junglebarry: This test uses the same setup as the test for `Views`.  I'll pare it back before merge.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
junglebarry: This will use a separate queue.  No effort yet on the mechanics of that.

junglebarry: (but I will need to change the code further to achieve that - the current form uses the same queue, but that code will be replaced.)

src/mongo/jobs/DiscoverOutdatedComposites.class.php
junglebarry: I think so, once we're settled on the query patterns.

src/mongo/delegates/Views.class.php
```diff
@@ -848,4 +877,15 @@ protected function getCollectionForViewSpec($viewSpecId)
...
+    
+    /**
+     * Regenerate a specific composite document from its spec and root CBD
+     * @param $specification
+     * @param $compositeCollection
+     * @param $rootCbd
+     * @return void
+     */
+    public function regenerateOne($specification, $compositeCollection, $rootCbd) {
+        $this->saveGeneratedView($specification, $compositeCollection, $rootCbd, $specification['from'], $this->defaultContext);
```
rsinger: A little unsure why this doesn't just use `generateView`?

src/mongo/ICompositeRegen.php
rsinger: Ok, that makes sense.

src/mongo/delegates/Views.class.php
rsinger: @junglebarry, what's the use case for instantiating a View without going through `Tripod`?  How would you deal with a `view_specifications` config where the various specs had different `to` properties?  We don't do this, currently, in views, but we do in tables.  I'm not sure I understand how you'd use a `View` object that wasn't rooted in a CBD originally.

src/mongo/delegates/Views.class.php
rsinger: I am all for optimizing ways to do this.  My comment wasn't necessarily a defense of the status quo, more just a rationale that we hadn't come up with/prioritized anything better.

Historically (and, I imagine, as you're currently seeing), there has been a lot of "HERE BE DRAGONS" when it comes to the composites functionality in Tripod.  A lot of reticence to pull at too many strings because, while we have a remit to fix broken windows, sometimes those windows are looking into the howling void of madness.  It's also a tremendous time sink not to mention a white-knuckle "how good is our test coverage _really_?" when you get into it.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
rsinger: A job for https://github.com/chrisboulton/php-resque-scheduler?

src/mongo/jobs/DiscoverOutdatedComposites.class.php
rsinger: Couldn't it also use something like `$exists: false` to find unversioned composites?

src/mongo/delegates/Views.class.php
```diff
@@ -202,8 +217,6 @@ public function getViewForResource($resource,$viewType,$context=null)
      */
     public function getViewForResources(Array $resources,$viewType,$context=null)
```
rsinger: I don't see anything in here (or `getViewForResource`) that regenerates the view if the retrieved document is from the wrong spec revision?

src/mongo/delegates/Views.class.php
junglebarry: All good.  I realise there's a bunch of stuff here that I'll hardly be the first to raise, and understand not wanting to optimise things without test coverage.  I'm going for things that I'm changing, and trying not to pull anything I need not change into the mix.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
junglebarry: @rsinger precisely.

src/mongo/Config.class.php
junglebarry: Agreed about traits - I've been longing for them.  I'm glad "this PR's lifetime" won't stretch that far, however :)

src/mongo/delegates/Views.class.php
junglebarry: No, @rsinger, you're quite right: there is nothing here to do that (in this PR).

@robotrobot and I had a conversation about the starting point, and he was concerned that if we didn't consider a scheduled job early, we might pick a poor representation for spec revisions.  I've started with a scheduled job, and will move to those later, perhaps in a follow-up PR.

src/mongo/delegates/Views.class.php
junglebarry: Because `generateView` doesn't only save; sometimes it queues for later.  When it does queue, it regenerates all composites for the CBD, not just the one we wish to target.  I will be using a queued approach in this PR eventually, but it should be targeted in what it regenerates.

junglebarry:
Thanks for your feedback so far.  I'd be glad of a discussion meeting, and would encourage everyone to consider the per-spec revision vs the global-revision+hash design before we do.

src/mongo/Config.class.php
junglebarry: Hmm... you're right, this is loose.

I picked up the terminology from the `_id` inside the composite document.  In the tests, for example, it looks like this:

```php
            "_id"=>array(
                _ID_RESOURCE=>"http://talisaspire.com/resources/filter1",
                _ID_CONTEXT=>'http://talisaspire.com/',
                "type"=>"v_resource_filter1"),
            _SPEC_KEY => array(
                _SPEC_TYPE => "v_resource_filter1",
                _SPEC_REVISION => 1
            ),
```

So `_id.type` and `_spec.type` would be identical.  I wasn't sure whether I could just add to the `_id` field, to save duplicating it.  In either case, I'll take advice on the naming here.

junglebarry: It'd probably make more sense if the example I used included the constant:

```php
define('_ID_TYPE','type');
```

that was the inspiration.

src/mongo/jobs/DiscoverOutdatedComposites.class.php
junglebarry: scala trolling

PR #106, opened 2016-08-19
90% TENT-1872 Add retry logic to tripod

rgubby:
If there is momentary blip in comms, the mongo php driver handles it (or is supposed to). What it doesn't handle however, is a step down for the primary on the repset.

In that instance, we get a 500 error with "No candidate servers found", when really it should back off and try again. The connection will be available in a few seconds once a new primary is elected and then the code can continue as per normal.

As far as i can tell, the mongo PHP driver does not support this so we're going to create something to do it for us.

You get 2 exceptions in Tripod when there are blips (step downs) in the rep set. The first one is when the rep set is down when you attempt to connect to it (`MongoConnectionException`). You also get one when the rep set is up when you connect, but down when you try to use it (`MongoCursorException`).

#### Questions
* ~~Is this the right approach?~~
* What is a reasonable back off strategy? I've currently got 30 seconds of retrying with no exponential back off - it's just a simple sleep for a second until it's ready. Maybe a back off for 0.5s intervals and do it 30 times, so 15s but if you got a connection earlier, it would resolve quicker rather than have to wait for a whole second. Do we need to make it an exponential back off?
* If no connection can be found at all, is it reasonable to throw the last exception that occurred?

#### Manual testing
1. `rs.stepDown()` on node 1 (primary)
1. `db.shutdownServer()` on node 1
1. Bring the service online again
1. `rs.stepDown()` on node 2 (now the primary)
1. `db.shutdownServer()` on node2
1. Bring the service online again

Whilst doing this, I ran an ab command to keep hitting a list:
```bash
ab -c 5 -t 100000 -v 3 "http://life.ac.uk/lists/fdta102.html"|grep "Response code"
```
That told me that we were consistently getting 200 responses.

Tailing the logs meant I could see the Exceptions coming through and clearing.

For future reference - in order to do stepDowns and shutdownServer commands, you need an arbiter running locally. You can do this with the following command (ymmv):
```bash
$ sudo mkdir /mnt/arb
$ sudo chown mongodb /mnt/arb
$ sudo mongod --port 3000 --dbpath /mnt/arb --smallfiles --replSet repSetName
```

#### Todo
* [x] Confirm this works correctly when doing a step down of the repset primary. 
* [x] 30% PR review
* [x] Tests
* [x] 90% PR review

### DIFF
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1767,6 +1767,7 @@ class Config
                 try {
                     $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
                 } catch (\MongoConnectionException $e) {
+                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
                     sleep(1);
                     $retries++;
                     $exception = $e;
@@ -1775,6 +1776,7 @@ class Config
             } while ($retries < self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) && $this->connections[$dataSource]>connected !== true ));
 
             if (!isset($this->connections[$dataSource])) {
+                self::getLogger()->warn("MongoConnectionException after " . self::CONNECTION_RETRIES . " attempts: " . $e->getMessage());
                 throw new \MongoConnectionException($exception);
             }
         }
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1762,10 +1762,16 @@ class Config
         {
             $retries = 1;
             $exception = null;
+            $connected = false;
 
             do {
                 try {
                     $this->connections[$dataSource] = $this->getMongoClient($ds['connection'], $connectionOptions);
+                    // Check if we are connected
+                    $connections = $this->connections[$dataSource]->getConnections();
+                    if (empty($connections) === false) {
+                        $connected = true;
+                    }
                 } catch (\MongoConnectionException $e) {
                     self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
                     sleep(1);
@@ -1773,7 +1779,7 @@ class Config
                     $exception = $e;
                 }
 
-            } while ($retries <= self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) || (isset($this->connections[$dataSource]) && $this->connections[$dataSource]->connected !== true)));
+            } while ($retries <= self::CONNECTION_RETRIES && $connected == false);
 
             if (!isset($this->connections[$dataSource])) {
                 self::getLogger()->warn("MongoConnectionException after " . $retries . " attempts (MAX:".self::CONNECTION_RETRIES."): " . $e->getMessage());
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1773,16 +1773,16 @@ class Config
                         $connected = true;
                     }
                 } catch (\MongoConnectionException $e) {
-                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
+                    self::getLogger()->error("MongoConnectionException attempt ".$retries.". Retrying...:" . $e->getMessage());
                     sleep(1);
                     $retries++;
                     $exception = $e;
                 }
 
-            } while ($retries <= self::CONNECTION_RETRIES && $connected == false);
+            } while ($retries <= self::CONNECTION_RETRIES && $connected === false);
 
             if (!isset($this->connections[$dataSource])) {
-                self::getLogger()->warn("MongoConnectionException after " . $retries . " attempts (MAX:".self::CONNECTION_RETRIES."): " . $e->getMessage());
+                self::getLogger()->error("MongoConnectionException failed after " . $retries . " attempts (MAX:".self::CONNECTION_RETRIES."): " . $e->getMessage());
                 throw new \MongoConnectionException($exception);
             }
         }
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -121,6 +121,7 @@ class Config
 
     const VALIDATE_MIN = 'MIN';
     const VALIDATE_MAX = 'MAX';
+    const CONNECTION_RETRIES = 30;
 
     /**
      * @var string
@@ -1759,7 +1760,23 @@ class Config
         }
         if(!isset($this->connections[$dataSource]))
         {
-            $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
+            $retries = 1;
+            $exception = null;
+
+            do {
+                try {
+                    $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
+                } catch (\MongoConnectionException $e) {
+                    sleep(1);
+                    $retries++;
+                    $exception = $e;
+                }
+
+            } while ($retries < self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) && $this->connections[$dataSource]>connected !== true ));
+
+            if (!isset($this->connections[$dataSource])) {
+                throw new \MongoConnectionException($exception);
+            }
         }
         return $this->connections[$dataSource];
     }
@@ -2113,3 +2130,4 @@ class Config
         return self::$logger;
     }
 }
+
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1773,7 +1773,7 @@ class Config
                     $exception = $e;
                 }
 
-            } while ($retries < self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) && $this->connections[$dataSource]>connected !== true ));
+            } while ($retries < self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) && $this->connections[$dataSource]->connected !== true ));
 
             if (!isset($this->connections[$dataSource])) {
                 self::getLogger()->warn("MongoConnectionException after " . self::CONNECTION_RETRIES . " attempts: " . $e->getMessage());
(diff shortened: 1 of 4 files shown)


src/mongo/Config.class.php
```diff
@@ -1759,7 +1760,23 @@ protected function getConnectionForDataSource($dataSource)
...
         if(!isset($this->connections[$dataSource]))
         {
-            $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
+            $retries = 1;
+            $exception = null;
+
+            do {
+                try {
+                    $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
+                } catch (\MongoConnectionException $e) {
```
rgubby: We need to specifically look for an error code here on the exception. If we had a situation where the cluster was offline, then it's going to lock up Apache processes for 30 seconds. This is going to lock up the servers and cause untold woe.

Better to be specific about the exception and try and capture a step down.

foomatty:
Agreed that initially we should put this retry in for 'No Candidate Servers Found' which is the typical message when Mongo connects to the repset but is unable to find a primary server.  Elections take anything from 15 to 60 seconds, so our retry logic should cover this, a back off strategy is fine but should have a max interval of 5 seconds.

For connection timeout errors, than just a retry might suffice.

For read timeouts (30 seconds) which we've seen a few of, it might be a good idea not to retry these as they are probably table scans.

foomatty:
@rsinger @malcyL some eyes on this please.

rsinger:
Have we confirmed that the latest PHP driver doesn't support this?  The possibility of waiting 30 secs. seems... problematic, but it's really hard to understand the implications, honestly.

foomatty:
@rsinger We're on mongo-1.6 of the latest of the legacy driver, there is no option within the driver to handle these exceptions.  The new driver (mongodb-1.1) doesn't appear to support such an option either, it is also PHP5.4 and is probably a larger code change.

There is a connection string option for connection timeout, this is most likely set to 60 seconds as it picks up the default PHP socket timeout.  We could lower this value, but has no effect with this PR.

I'd like to get this pushed through, we have some repset fail overs to do for upgrades, and maintenance windows in this area really slow us down.

rgubby:
I've looked at the error codes that we get - and if we use a repset, we always get the same error code + message: "No candidate servers found" - so i have left it with no check in as it's always going to be performing this code.

The cursor exception always gives us the same exception as well, so have left that as it is too.

src/mongo/Config.class.php
kiyanwang: Where you able to find an appropriate error code.

src/mongo/Config.class.php
```diff
@@ -121,6 +121,7 @@ class Config
 
     const VALIDATE_MIN = 'MIN';
     const VALIDATE_MAX = 'MAX';
+    const CONNECTION_RETRIES = 30;
```
kiyanwang: so the longest we want it to retry for is 30 seconds given we have a 30 second cut off in cloudfront for https tenancies. which makes sense.

src/mongo/base/DriverBase.class.php
```diff
@@ -183,22 +183,43 @@ protected function fetchGraph($query, $type, $collection=null,$includeProperties
...
                 }
+                $cursorSuccess = true;
+            } catch (\MongoCursorException $e) {
+                self::getLogger()->warn("MongoCursorException: " . $e->getMessage());
+                sleep(1);
+                $retries++;
+                $exception = $e;
             }
-            $graph->add_tripod_array($result);
+        } while ($retries < 30 && $cursorSuccess === false);
```
kiyanwang: seems reasonable.   was going to make the point that it makes sense to retry if the nature of the cursor exception is something you know you can recover from. but i dont think you can distinguish so a brute retry is probably all you can do here.

src/mongo/Config.class.php
```diff
@@ -1759,7 +1760,25 @@ protected function getConnectionForDataSource($dataSource)
...
+                try {
+                    $this->connections[$dataSource] = new \MongoClient($ds['connection'], $connectionOptions);
+                } catch (\MongoConnectionException $e) {
+                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
+                    sleep(1);
+                    $retries++;
+                    $exception = $e;
+                }
+
+            } while ($retries < self::CONNECTION_RETRIES && (!isset($this->connections[$dataSource]) && $this->connections[$dataSource]>connected !== true ));
```
kiyanwang:
```php
$this->connections[$dataSource]>connected !== true
````

is that correct?   `>connected` ?

Comment (kiyanwang) [inline src/mongo/Config.class.php]: shouldnt it be `->connected !== true`

**Comment by kiyanwang** [comment]
In terms of the questions you ask in the description for the pr:

The backoff strategy is something I think we can optimise after we know the approach actually resolves the problem.  The priority for me is that the retry works; and we can see that its working. Once we we know it is, we could then look at optimising the back off.

File: src/mongo/Config.class.php (hunk 1 above)
Comment (rgubby) [inline src/mongo/Config.class.php]: There is one error code here - 71 - but you get that for every mongo connection exception so i didn't track for one specifically.

File: src/mongo/base/DriverBase.class.php (hunk 3 above)
Comment (rgubby) [inline src/mongo/base/DriverBase.class.php]: I can't distinguish between diff types of exception so had to brute force it.

File: src/mongo/Config.class.php [hunk 5]
```diff
@@ -1759,12 +1760,47 @@ protected function getConnectionForDataSource($dataSource)
...
+                        $connected = true;
+                    }
+                } catch (\MongoConnectionException $e) {
+                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
+                    sleep(1);
+                    $retries++;
+                    $exception = $e;
+                }
+
+            } while ($retries <= self::CONNECTION_RETRIES && $connected == false);
```
kiyanwang: `=== false` ?

src/mongo/Config.class.php
```diff
@@ -1759,12 +1760,47 @@ protected function getConnectionForDataSource($dataSource)
...
 
     /**
+     * Create a Mongo Client - used for mocking
+     * @param string $connection
+     * @param array $connectionOptions
+     * @return \MongoClient
+     */
+    protected function getMongoClient($connection, $connectionOptions)
+    {
+        return new \MongoClient($connection, $connectionOptions);
```
kiyanwang: do you want a new mongo client every time? or should this be a lazy load? what (if any ) is the overhead of creating a new client each time?

src/mongo/Config.class.php
```diff
@@ -1759,12 +1760,47 @@ protected function getConnectionForDataSource($dataSource)
...
+                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
+                    sleep(1);
+                    $retries++;
+                    $exception = $e;
+                }
+
+            } while ($retries <= self::CONNECTION_RETRIES && $connected == false);
+
+            if (!isset($this->connections[$dataSource])) {
+                self::getLogger()->warn("MongoConnectionException after " . $retries . " attempts (MAX:".self::CONNECTION_RETRIES."): " . $e->getMessage());
```
malcyL: Error rather than warning?

src/mongo/Config.class.php
```diff
@@ -1759,12 +1760,47 @@ protected function getConnectionForDataSource($dataSource)
...
+            do {
+                try {
+                    $this->connections[$dataSource] = $this->getMongoClient($ds['connection'], $connectionOptions);
+                    // Check if we are connected
+                    $connections = $this->connections[$dataSource]->getConnections();
+                    if (empty($connections) === false) {
+                        $connected = true;
+                    }
+                } catch (\MongoConnectionException $e) {
+                    self::getLogger()->warn("MongoConnectionException: " . $e->getMessage());
```
malcyL: I'd like to see a little more information in the log.

Maybe something like:
```
"Connection attempt 1 failed. Retrying..."
"Connection attempt 2 failed. Retrying..."
"Connection attempt 30 failed. Max Retry reached."
```

This would give us more information on how often this was happening, how well we recovered.

src/mongo/base/DriverBase.class.php
```diff
@@ -183,22 +183,43 @@ protected function fetchGraph($query, $type, $collection=null,$includeProperties
...
+                        {
+                            // regenerate!
+                            $this->generateView($result['_id']['type'],$result['_id']['r']);
+                        }
+                    }
+                    $graph->add_tripod_array($result);
                 }
+                $cursorSuccess = true;
+            } catch (\MongoCursorException $e) {
+                self::getLogger()->warn("MongoCursorException: " . $e->getMessage());
```
malcyL: Same comment re the logging as above.

src/mongo/base/DriverBase.class.php
```diff
@@ -183,22 +183,43 @@ protected function fetchGraph($query, $type, $collection=null,$includeProperties
...
+                sleep(1);
+                $retries++;
+                $exception = $e;
             }
-            $graph->add_tripod_array($result);
+        } while ($retries <= Config::CONNECTION_RETRIES && $cursorSuccess === false);
+
+        if ($cursorSuccess === false) {
+            self::getLogger()->warn("MongoCursorException after " . $retries . " attempts (MAX:".Config::CONNECTION_RETRIES."): " . $e->getMessage());
```
malcyL: warn -> error?

malcyL:
:+1:

rgubby:
Even though GitHub isn't showing the changes, I've addressed the issues above.

rgubby:
For reference - here is some log output:
```bash
[Fri Sep 16 13:12:50 2016] [life.ac.uk] [57dbe1bd6d793] [ERROR] [Session Id: nhvdn69qsq4jo85qf2hdrfu116] - MongoCursorException attempt 1. Retrying...:localhost:27018: Remote server has closed the connection []
```

Issue #109, opened 2016-09-26
Upgrade to latest series PHP driver


robotrobot:
The official mongo driver for PHP is now in two parts - the underlying driver, which now mirrors the MongoDB wire protocol more closely, and a PHP Mongo lib, which mirrors closely the old mongo.so PECL.

When this ticket is fixed, tripod will work against the new Mongo lib (brought in via composer) and the underlying mongodb.so which allows us to target PHP > 5.4, PHP7 and HHVM, as well as make use of Mongo 3.x features.

This will be a major semver bump and a breaking change.

rgubby:
Resolved with https://github.com/talis/tripod-php/pull/111

PR #111, opened 2016-09-27
Issue #109 Upgrade to latest PHP driver

rgubby:
This is to support the new mongo driver for PHP.

The PHP drivers implement a different API, so all chainable calls such as `->find()->sort()->limit()` have been replaced with another format. Read preferences have changed (how you set them), so have some of the function calls. An upgrade guide can be found here: https://mongodb.github.io/mongo-php-library/upgrade-guide/

There are some requirement changes:
- Mongo 2.6.x and above. This has changed from requiring Mongo 2.x - the reason for this is the way an `updateOne` or `replaceOne` function works with the current functionality with the new PHP mongo library. If we want to support older versions of Mongo, we'll need to change some functionality.
- PHP 5.4+. The MongoDB PHP library only works with PHP 5.4 and above - so we're tied to this. Also - we've started using some PHP 5.4 and above functionality (array notation for a start).
- Added compatibility for PHP 7.0 (7.1 has unit test failures so I haven't added that).

A preview modification for Tripod involved retrying connections when stepping down and removing servers from the repset. I've gone through the same testing process with this new branch as I did on that PR. All works as expected.
#### Todo
- [x] Decide if we're going to support BSONDocuments in MongoGraph or leave it as arrays
- [x] `db->getLastError()` does not work anymore - find a solution
- [x] Look into how to set read preferences correctly rather than how we were doing it in the old driver
- [x] Transaction log and document locking to work
- [x] Replace all instances of Mongo\* - not touched the scripts/utils yet so there are some instances.
- [x] Look into how to manage a successful connection without doing a PING command
- [x] Travis support for new mongo driver - remove the old one
- [x] Can you define a mongo version in Travis?
- [x] Check docblocks and type hinting are correct
- [x] Check for all instances of Mongo*

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     "require": {
         "semsol/arc2": "v2.2.4",
         "chrisboulton/php-resque": "dev-master#98fde571db008a8b48e73022599d1d1c07d4a7b5",
-        "monolog/monolog" : "~1.13"
+        "monolog/monolog" : "~1.13",
+        "mongodb/mongodb": "^1.0.0"
     },
     "require-dev": {
         "phpunit/phpunit": "4.1.*"
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -1,6 +1,15 @@
 <?php
 
 namespace Tripod\Mongo;
+
+use \MongoDB\Client;
+use \MongoDB\Database;
+use \MongoDB\Collection;
+use \MongoDB\Driver\ReadPreference;
+use \MongoDB\Driver\Command;
+use \MongoDB\Driver\Manager;
+use \MongoDB\Driver\Exception;
+
 /**
  * Holds the global configuration for Tripod
  */
@@ -1716,13 +1725,13 @@ class Config
     }
 
     /**
-     * @param $storeName
+     * @param string $storeName
      * @param string|null $dataSource
      * @param string $readPreference
      * @throws \Tripod\Exceptions\ConfigException
-     * @return \MongoDB
+     * @return Database
      */
-    public function getDatabase($storeName, $dataSource = null, $readPreference = \MongoClient::RP_PRIMARY_PREFERRED)
+    public function getDatabase($storeName, $dataSource = null, $readPreference = ReadPreference::RP_PRIMARY_PREFERRED)
     {
         if(!isset($this->dbConfig[$storeName]))
         {
@@ -1735,15 +1744,17 @@ class Config
         }
 
         $client = $this->getConnectionForDataSource($dataSource);
-        $db = $client->selectDB($this->dbConfig[$storeName]['database']);
-        $db->setReadPreference($readPreference);
+        $db = $client->selectDatabase($this->dbConfig[$storeName]['database'], array(
+            'readPreference' => new ReadPreference($readPreference)
+        ));
         return $db;
     }
 
     /**
      * @param string $dataSource
-     * @return \MongoClient
+     * @return Client
      * @throws \Tripod\Exceptions\ConfigException
+     * @throws ConnectionTimeoutException
      */
     protected function getConnectionForDataSource($dataSource)
     {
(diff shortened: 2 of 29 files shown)


composer.json
```diff
@@ -20,7 +20,8 @@
     "require": {
         "semsol/arc2": "v2.2.4",
         "chrisboulton/php-resque": "dev-master#98fde571db008a8b48e73022599d1d1c07d4a7b5",
-        "monolog/monolog" : "~1.13"
+        "monolog/monolog" : "~1.13",
```
rgubby: I don't think they should - libraries should have fuzzy versions whereas apps lock them down.

src/mongo/Config.class.php
```diff
@@ -1762,50 +1773,49 @@ protected function getConnectionForDataSource($dataSource)
...
                     $exception = $e;
                 }
 
-            } while ($retries <= self::CONNECTION_RETRIES && $connected === false);
+            } while ($retries <= self::CONNECTION_RETRIES);
 
             if (!isset($this->connections[$dataSource])) {
                 self::getLogger()->error("MongoConnectionException failed after " . $retries . " attempts (MAX:".self::CONNECTION_RETRIES."): " . $e->getMessage());
-                throw new \MongoConnectionException($exception);
+                throw new ConnectionTimeoutException($exception);
```
rgubby: No nasty ping - if the connection object is created successfully (no exception thrown), then i'm breaking on the line afterwards to get out of the do..while loop.

src/mongo/delegates/Updates.class.php
```diff
@@ -229,19 +235,26 @@ public function saveChanges(
     protected function setReadPreferenceToPrimary()
     {
         // Set db preference
-        $dbPref = $this->getDatabase()->getReadPreference();
-        if($dbPref['type'] !== \MongoClient::RP_PRIMARY){
-            $this->originalDbReadPreference = $this->db->getReadPreference();
-            $tagsets = (isset($dbPref['tagsets']) ? $dbPref['tagsets'] : array());
-            $this->db->setReadPreference(\MongoClient::RP_PRIMARY, $tagsets);
+        /** @var ReadPreference $dbReadPref */
+        $dbReadPref = $this->getDatabase()->__debugInfo()['readPreference'];
```
kiyanwang: Interesting that debug info is a magic method here? can we rely on this? sorry what I mean is, clearly its meant to be used only for debugging, which means it could change massively btw versions of the driver, but what seem to be reliant on it?

kiyanwang:
did you check to see if the scripts in `scripts/mongo` are affected by these changes?

src/mongo/delegates/Tables.class.php
```diff
@@ -362,8 +371,9 @@ public function deleteTableRowsByTableId($tableId) {
             return;
         }
 
+        $query = array("_id.type"=>$tableId);
         $this->config->getCollectionForTable($this->storeName, $tableId)
-            ->remove(array("_id.type"=>$tableId), array('fsync'=>true));
+            ->deleteMany($query);
```
kiyanwang: do you need to inspect the deleteResult object to verify that it actually worked or did what you expected it to?

src/mongo/delegates/Updates.class.php
rgubby: It's not implemented currently - so at some point i imagine it will be and we can remove that. At which point we'll be bumping up a change of the driver and we'll be using new functionality. We should probably raise an issue to track that though.

rgubby:
> did you check to see if the scripts in scripts/mongo are affected by these changes? in principle the changes should be transparent but is it worth checking?

Yes - i checked - there was an amendment made to just one of the scripts but no others - they use Tripod methods to do anything.

src/mongo/Config.class.php
```diff
@@ -1762,50 +1773,49 @@ protected function getConnectionForDataSource($dataSource)
...
                 try {
-                    $this->connections[$dataSource] = $this->getMongoClient($ds['connection'], $connectionOptions);
-                    // Check if we are connected
-                    $connections = $this->connections[$dataSource]->getConnections();
-                    if (empty($connections) === false) {
-                        $connected = true;
-                    }
-                } catch (\MongoConnectionException $e) {
-                    self::getLogger()->error("MongoConnectionException attempt ".$retries.". Retrying...:" . $e->getMessage());
+                    $connectionString = $ds['connection'] . '?' . http_build_query($connectionOptions);
```
rgubby: It's validated in the `loadConfig` function. `$connectionOptions` is defensive just above this code - it only adds entries if things exist. And the `$ds['connection']` string HAS to be present. If it doesn't a `ConfigException` exception would be thrown much earlier in the execution.

src/mongo/delegates/Tables.class.php
rgubby: I've tracked the result of this and logged out if there was an issue acknowledging the delete/update.

PR #114, opened 2016-11-08
70% Introduced resque worker to create all indexes

### LINKED ISSUE #113: Move index creation into a background job
At the moment we create indexes for CBD, tables, views and search documents every time we insert a new document into the relevant collections. this is both inefficient and leads to inconsistency.

We need to remove the code that is doing this from within each of the delegates and refactor that into a background resque job that can be scheduled to run as (in)frequently as needed

kiyanwang:
- added constants for resque job

ensureIndexes now creates all required indexes

IndexUtils->ensureIndexes has been updated
- this method now ensures ALL tripod required indexes for Tables
- this method now ensures ALL tripod required indexes for Views
- this method now ensures ALL tripod required indexes for Search Docs

Removed hardcoded index creation from delegates

on every insert/update of a composite we were ensuring a set of hardcoded
indexes, in addition to ensuring ALL indexes specified in the compose spec i.e.
in the View/Table/SearchDoc Specification.  All of this code has been removed,
in favor of a more coherent approach via background jobs facilitated by the
IndexUtils->ensureIndexes method.

removed redundant tests from MongoTripodTablesTest

- these tests contained expectations around calls to the Tables->ensureIndex
  method which now, no longer exists.

Also added tests for IndexUtils which previously had no test coverage.
This class does the actual work now around creating indexes

Fixes #113


src/mongo/jobs/EnsureIndexes.class.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @return \Tripod\Mongo\IndexUtils
+     */
+    protected function getIndexUtils()
+    {
+        return new \Tripod\Mongo\IndexUtils();
+    }
+
+    public function createJob($storeName, $reindex, $queueName=null)
```
rsinger: Docblock

src/mongo/util/IndexUtils.class.php
```diff
@@ -65,10 +65,14 @@ public function ensureIndexes($reindex=false,$storeName=null,$background=true)
             // Index views
             foreach($config->getViewSpecifications($storeName) as $viewId=>$spec)
             {
-                $collection = Config::getInstance()->getCollectionForView($storeName, $viewId);
+                $collection = $config->getCollectionForView($storeName, $viewId);
                 if($collection)
                 {
-                    $indexes = array(array("_id.type"=>1));
+                    $indexes = array(
+                        array("_id.r" =>1, "_id.c"=>1, "_id.type"=>1),
```
rsinger: We have constants for these keys, don't we?

src/mongo/util/IndexUtils.class.php
```diff
@@ -92,10 +96,14 @@ public function ensureIndexes($reindex=false,$storeName=null,$background=true)
             // Index table rows
             foreach($config->getTableSpecifications($storeName) as $tableId=>$spec)
             {
-                $collection = Config::getInstance()->getCollectionForTable($storeName, $tableId);
+                $collection = $config->getCollectionForTable($storeName, $tableId);
                 if($collection)
                 {
-                    $indexes = array(array("_id.type"=>1));
+                    $indexes = array(
+                        array("_id.r" =>1, "_id.c"=>1, "_id.type"=>1),
```
rsinger: As above, re: constants

src/mongo/jobs/EnsureIndexes.class.php
rsinger: In `IndexUtils`, storename is optional - how would you schedule an index creation for all stores?

rsinger:
Overall, this looks sane.  I don't know that I quite understand what prompts this job getting queued, though.

kiyanwang:
@rsinger the jobs will be queued via the setup console. which im changing at the moment.

rsinger:
Ok, but let's not forget that Tripod is OSS.  Not that I think we need to specifically cater to the use case that somebody somewhere might be using it, but a 3rd party isn't going to have the option of doing this in setup console :D

rsinger:
But that does harken back to my disconnect between the submit job expecting a store and the actual perform job being able to deal with all stores.

.gitignore
```diff
@@ -2,11 +2,10 @@
 composer.phar
 output
 vendor
-
+tags
```
rgubby: for my benefit - what are "tags" ?

src/mongo/jobs/EnsureIndexes.class.php
```diff
@@ -0,0 +1,92 @@
...
+        {
+            $queueName = \Tripod\Mongo\Config::getEnsureIndexesQueueName();
+        }
+        elseif(strpos($queueName, \Tripod\Mongo\Config::getEnsureIndexesQueueName()) === false)
+        {
+            $queueName = \Tripod\Mongo\Config::getEnsureIndexesQueueName() . '::' . $queueName;
+        }
+
+        $data = array(
+            self::STORENAME_KEY=>$storeName,
```
rgubby: Can we get some extra spaces around the `=>` ?

src/mongo/jobs/EnsureIndexes.class.php
```diff
@@ -0,0 +1,92 @@
...
+        catch(\Exception $e)
+        {
+            $this->getStat()->increment(MONGO_QUEUE_ENSURE_INDEXES_FAIL);
+            $this->errorLog("Caught exception in ".get_class($this).": ".$e->getMessage());
+            throw $e;
+        }
+    }
+
+    /**
+     * @param string $storeName
```
rgubby: Can we get a docblock here to explain that the function here is the one that creates the job.

src/mongo/jobs/EnsureIndexes.class.php
```diff
@@ -0,0 +1,92 @@
...
+            $timer->start();
+
+            $this->validateArgs();
+
+            \Tripod\Mongo\Config::setConfig($this->args[self::TRIPOD_CONFIG_KEY]);
+
+            $this->getIndexUtils()->ensureIndexes(
+                $this->args[self::REINDEX_KEY],
+                $this->args[self::STORENAME_KEY],
+                true // always create indexes in the background
```
rgubby: Would you ever want to create them in the foreground?

src/mongo/util/IndexUtils.class.php
```diff
@@ -65,10 +65,14 @@ public function ensureIndexes($reindex=false,$storeName=null,$background=true)
...
             {
-                $collection = Config::getInstance()->getCollectionForView($storeName, $viewId);
+                $collection = $config->getCollectionForView($storeName, $viewId);
                 if($collection)
                 {
-                    $indexes = array(array("_id.type"=>1));
+                    $indexes = array(
+                        array(_ID_KEY.'.'._ID_RESOURCE=>1, _ID_KEY.'.'._ID_CONTEXT=>1, _ID_KEY.'.'._ID_TYPE=>1),
+                        array(_ID_KEY.'.'._ID_TYPE=>1),
+                        array("value."._IMPACT_INDEX=>1)
```
rgubby: is there no constant for `"value.impactIndex"`?  Surprised we haven't got anything already!

rgubby: Also - spaces around `=>`

kiyanwang:
@rsinger theres also a script in tripod/scripts that acts a cli tool for queuing these jobs

src/mongo/util/IndexUtils.class.php
```diff
@@ -115,6 +123,39 @@ public function ensureIndexes($reindex=false,$storeName=null,$background=true)
...
+                                "background"=>$background
+                            )
+                        );
+                    }
+                }
+            }
         }
     }
+
+    protected function getConfig()
```
rgubby: docblock

test/unit/mongo/EnsureIndexesTest.php
```diff
@@ -0,0 +1,265 @@
...
+        );
+
+        //create mock job
+        $job = $this->getMockBuilder('\Tripod\Mongo\Jobs\EnsureIndexes')
+            ->setMethods(array('warningLog', 'enqueue'))
+            ->getMock();
+
+        $e = new Exception("Connection to Redis failed after 1 failures.Last Error : (0) php_network_getaddresses: getaddrinfo failed: nodename nor servname provided, or not known");
+
+        $job->expects($this->any())->method("enqueue")->will($this->throwException($e));
```
rgubby: Do you want to be more specific over the number of times this mocked method will be called (same goes for the other ones below)

src/mongo/jobs/EnsureIndexes.class.php
rgubby: We support them being created in the background/foreground in the ensureIndexes function - maybe we should support creating them in the job, even if we default back to background. Or maybe hard coding it to `true` here is good enough as we may always want to create it in the BG.

test/unit/mongo/IndexUtilsTest.php
```diff
@@ -0,0 +1,654 @@
...
+    {
+        // create index is called 4 times, each time with a different set of
+        // params that we know.
+        // a) one custom index is created based on the collection specification
+        // b) three internal indexes are always created
+        $mockCollection->expects($this->exactly(4))
+            ->method('createIndex')
+            ->withConsecutive(
+                array(array('rdf:type.u' => 1), array('name'=>'rdf_type', 'background'=>$background)),
+                array(array('_id' =>1 , '_lockedForTrans'=>1), array('name'=>'_lockedForTransIdx', 'background'=>$background)),
```
rgubby: constants here instead of `_id`?

test/unit/mongo/IndexUtilsTest.php
```diff
@@ -0,0 +1,654 @@
...
+    /**
+     * Returns tripod config to use on with the IndexUtils object
+     * This is a minimal config used to assert what should happen when ensuring
+     * indexes for a CBD collection
+     *
+     * @param PHPUnit_Framework_MockObject_MockObject mock Tripod Config object
+     */
+    protected function setConfigForCBDIndexes($mockConfig)
+    {
+
```
rgubby: extra space in function

test/unit/mongo/IndexUtilsTest.php
```diff
@@ -0,0 +1,654 @@
...
+            "collection"=>"transaction_log",
+            "data_source"=>"tlog"
+        );
+
+        $mockConfig->loadConfig($config);
+    }
+
+    protected function setConfigForViewIndexes($mockConfig)
+    {
+
```
rgubby: extra space in function

test/unit/mongo/IndexUtilsTest.php
```diff
@@ -0,0 +1,654 @@
...
+            "data_source"=>"tlog"
+        );
+
+        $mockConfig->loadConfig($config);
+    }
+
+    protected function setConfigForSearchDocIndexes($mockConfig)
+    {
+
+        // minimal config to verify that
```
rgubby: to verify that ...

.gitignore
kiyanwang: oh sorry, in vim I use a plugin called exhuberant ctags, it creates an index on disk for all symbols in the project which allows me to do refactoring, jump to source etc.  makes vim feel like an IDE.  its the equivalent of the .idea directory you have with phpstorm

PR #119, opened 2017-02-23
Cast UTCDateTime to a DateTime object for comparison

rgubby:
When using a conditional in a table spec that compares 2 date objects, the comparison fails because the 2 objects aren't in a state that can be compared. 

`UTCDateTime` cannot be compared using comparison operators `UTCDateTime`. A `DateTime` object however, can be.


rsinger:
👍  Seems good to me

PR #121, opened 2017-11-02
Use docker image rather than (non-working) .debs

rsinger:
mongodb-org 2.6.* is basically impossible to find as a .deb, so enable it via docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ php:
 - 5.4
 services:
 - redis
+- docker
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,8 +7,10 @@ php:
 - 5.4
 services:
 - redis
+- docker
 before_install:
-- sudo apt-get install -y mongodb-org=2.6.9 mongodb-org-server=2.6.9 mongodb-org-shell=2.6.9 mongodb-org-mongos=2.6.9 mongodb-org-tools=2.6.9
+- docker pull rossfsinger/mongo-2.6.12
```
rgubby: Can we add this to the Talis account? If you don't have permissions, can @omarqureshi add you?

rsinger: I don't have permissions to do that, I don't think.  `talis` wasn't an option for me when I created it.

PR #128, opened 2018-08-23
Batch generate composites

rsinger:
Presently, an individual job is queued for each document that needs composite generation.  This causes an unnecessary bottleneck by queuing up (potentially) many ApplyOperation Resque jobs to update one resource: Resque needs to contact Redis, then make a Mongo connection, do validation, etc. perform a very small update, then close the mongo connection, etc.

A much more efficient way to do this is to *batch* the ApplyOperation jobs with, e.g., 100 table row operations.  It's already designed to work this way, we just aren't utilizing it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
 mongo:
-  image: mongo/2.6.12:latest
+  image: rossfsinger/mongo-2.6.12:latest
   ports:
       - "27017:27017"
--- a/src/mongo/base/CompositeBase.class.php
+++ b/src/mongo/base/CompositeBase.class.php
@@ -12,70 +12,67 @@ abstract class CompositeBase extends \Tripod\Mongo\DriverBase implements \Tripod
      * @var \Tripod\Mongo\Jobs\ApplyOperation
      */
     protected $applyOperation;
+
+    protected $batchSize = 1;
+
     /**
      * Returns an array of ImpactedSubjects based on the subjects and predicates of change
      * @param array $subjectsAndPredicatesOfChange
      * @param string $contextAlias
      * @return \Tripod\Mongo\ImpactedSubject[]
      */
-    public function getImpactedSubjects(Array $subjectsAndPredicatesOfChange,$contextAlias)
+    public function getImpactedSubjects(array $subjectsAndPredicatesOfChange, $contextAlias)
     {
-        $candidates = array();
-        $filter = array();
-        $subjectsToAlias = array();
-        foreach(array_keys($subjectsAndPredicatesOfChange) as $s){
+        $candidates = [];
+        $filter = [];
+        $subjectsToAlias = [];
+        foreach (array_keys($subjectsAndPredicatesOfChange) as $s) {
             $resourceAlias = $this->labeller->uri_to_alias($s);
             $subjectsToAlias[$s] = $resourceAlias;
             // build $filter for queries to impact index
-            $filter[] = array(_ID_RESOURCE=>$resourceAlias,_ID_CONTEXT=>$contextAlias);
+            $filter[] = [_ID_RESOURCE=>$resourceAlias,_ID_CONTEXT=>$contextAlias];
         }
-        $query = array(_ID_KEY=>array('$in'=>$filter));
-        $docs = $this->getCollection()->find($query, array(
-            'projection' => array(_ID_KEY=>true, 'rdf:type'=>true)
-        ));
+        $query = [_ID_KEY=> ['$in' => $filter]];
+        $docs = $this->getCollection()->find(
+            $query,
+            ['projection' => [_ID_KEY => true, 'rdf:type' => true]]
+        );
 
         $types = $this->getTypesInSpecifications();
 
-        if($this->getCollection()->count($query) !== 0 ) {
-            foreach($docs as $doc)
-            {
+        if ($this->getCollection()->count($query) !== 0) {
+            foreach ($docs as $doc) {
                 $docResource = $doc[_ID_KEY][_ID_RESOURCE];
                 $docContext  = $doc[_ID_KEY][_ID_CONTEXT];
                 $docHash     = md5($docResource.$docContext);
 
-                $docTypes = array();
-                if(isset($doc["rdf:type"])) {
-                    if(isset($doc["rdf:type"][VALUE_URI])){
-                        $docTypes[] = $doc["rdf:type"][VALUE_URI];
+                $docTypes = [];
+                if (isset($doc['rdf:type'])) {
+                    if (isset($doc['rdf:type'][VALUE_URI])) {
+                        $docTypes[] = $doc['rdf:type'][VALUE_URI];
                     } else {
-                        foreach($doc["rdf:type"] as $t){
-                            if(isset($t[VALUE_URI]))
-                            {
+                        foreach ($doc['rdf:type'] as $t) {
+                            if (isset($t[VALUE_URI])) {
                                 $docTypes[] = $t[VALUE_URI];
                             }
                         }
                     }
                 }
 
-                $currentSubjectProperties = array();
-                if(isset($subjectsAndPredicatesOfChange[$docResource]))
-                {
+                $currentSubjectProperties = [];
+                if (isset($subjectsAndPredicatesOfChange[$docResource])) {
                     $currentSubjectProperties = $subjectsAndPredicatesOfChange[$docResource];
-                }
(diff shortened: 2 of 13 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,4 @@
 mongo:
-  image: mongo/2.6.12:latest
+  image: rossfsinger/mongo-2.6.12:latest
```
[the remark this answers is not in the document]
rsinger: Yep.  I'll figure that out later.

src/mongo/base/CompositeBase.class.php
```diff
@@ -12,111 +14,105 @@
      * @var \Tripod\Mongo\Jobs\ApplyOperation
      */
     protected $applyOperation;
+
+    protected $batchSize = 1;
```
rgubby: Is this used? I can see a `$batchSize` in config - and when you get the batch size you are getting it from the config object. I can't see this being used at all.

src/mongo/delegates/SearchIndexer.class.php
```diff
@@ -13,6 +13,8 @@
 use Tripod\Mongo\JobGroup;
```
rgubby: Now that you've replaced the `new JobGroup` with `$this->getJobGroup()`, this line can go.

src/mongo/delegates/SearchIndexer.class.php
```diff
@@ -44,13 +46,7 @@ public function __construct(\Tripod\Mongo\Driver $tripod, $readPreference = Read
...
         $this->stat = $tripod->getStat();
         $this->config = $this->getConfigInstance();
-        $provider = $this->config->getSearchProviderClassName($this->tripod->getStoreName());
-
-        if (class_exists($provider)) {
-            $this->configuredProvider = new $provider($this->tripod);
-        } else {
-            throw new \Tripod\Exceptions\SearchException("Did not recognise Search Provider, or could not find class: $provider");
-        }
+        $this->setSearchProvider($this->tripod, $this->config);
```
rgubby: There is a mismatch in what is being passed around.

`setSearchProvider` expects the 2nd param to be a `Config`, but `$this->config` is an `IConfigInstance`. I think it's just a docblock error but it would be good to resolve

src/mongo/delegates/SearchIndexer.class.php
```diff
@@ -302,14 +300,37 @@ protected function getSearchDocumentGenerator(Collection $collection, $context )
...
         }
         return $output;
     }
+
+    /**
+     * For mocking
+     *
+     * @param Driver $tripod Mongo Tripod Driver
+     * @param Config $config Mongo Tripod ConfigInstance
+     * @return void
```
rgubby: Missing `@throws \Tripod\Exceptions\SearchException`

src/mongo/delegates/SearchIndexer.class.php
```diff
@@ -234,7 +224,11 @@ public function generateSearchDocuments(
                     array($searchDocumentType)
                 );
 
-                $this->getApplyOperation()->createJob(array($subject), $queueName, $jobOptions);
+                $subjects[] = $subject;
```
rgubby: Is this block worth a comment to explain what it's doing now that it's not as simple as 1 create job for every item in the for loop?

src/mongo/delegates/Tables.class.php
```diff
@@ -515,32 +509,37 @@ public function generateTableRows($tableType, $resource = null, $context = null,
...
+        $count = $this->getConfigInstance()->getCollectionForCBD($this->storeName, $from)->count($filter);
+        $docs = $this->getConfigInstance()
+            ->getCollectionForCBD($this->storeName, $from)
+            ->find($filter, ['maxTimeMS' => 1000000]);
 
         $jobOptions = [];
+        $subjects = [];
         if ($queueName && !$resource && ($this->stat || !empty($this->statsConfig))) {
             $jobOptions['statsConfig'] = $this->getStatsConfig();
-            $jobGroup = new JobGroup($this->storeName);
```
rgubby: You can remove `use \Tripod\Mongo\JobGroup;` from the top now

src/mongo/delegates/Tables.class.php
```diff
@@ -554,8 +553,11 @@ public function generateTableRows($tableType, $resource = null, $context = null,
                     $from,
                     array($tableType)
                 );
-
-                $this->getApplyOperation()->createJob(array($subject), $queueName, $jobOptions);
+                $subjects[] = $subject;
```
rgubby: Just a thought (and not a blocker for this), but is it worth putting this logic in a class higher up the food chain / separate class/util rather than repeated in each composite?

src/mongo/delegates/Views.class.php
```diff
@@ -454,17 +454,18 @@ public function generateView($viewId, $resource = null, $context = null, $queueN
...
             'maxTimeMS' => $this->getConfigInstance()->getMongoCursorTimeout()
         ));
 
 
 
         $jobOptions = [];
+        $subjects = [];
         if ($queueName && !$resource) {
             $jobOptions['statsConfig'] = $this->getStatsConfig();
-            $jobGroup = new JobGroup($this->storeName);
```
rgubby: You can remove `use Tripod\Mongo\JobGroup;` from the top now

src/mongo/delegates/Views.class.php
```diff
@@ -454,17 +454,18 @@ public function generateView($viewId, $resource = null, $context = null, $queueN
         }
 
         // @todo Change this to a command when we upgrade MongoDB to 1.1+
-        $count = $this->config->getCollectionForCBD($this->storeName, $from)->count($filter);
-        $docs = $this->config->getCollectionForCBD($this->storeName, $from)->find($filter, array(
+        $count = $this->getConfigInstance()->getCollectionForCBD($this->storeName, $from)->count($filter);
```
rgubby: 👍

test/unit/mongo/MongoTripodSearchIndexerTest.php
```diff
@@ -499,4 +499,85 @@ public function testSavingMultipleNewEntitiesResultsInOneImpactedSubject()
...
+        $docs = [];
+
+        $configOptions = json_decode(file_get_contents(__DIR__ . '/data/config.json'), true);
+
+        for ($i = 0; $i < $count; $i++) {
+            $docs[] = ['_id' => ['r' => 'tenantLists:batch' . $i, 'c' => 'tenantContexts:DefaultGraph']];
+        }
+
+        $fakeCursor = new ArrayIterator($docs);
+        /** @var \PHPUnit_Framework_MockObject_MockObject|TripodTestConfig $config */
```
rgubby: `$config` -> `$configInstance`

test/unit/mongo/MongoTripodTablesTest.php
```diff
@@ -303,6 +316,74 @@ public function testGenerateTableRows()
...
+        $docs = [];
+
+        $configOptions = json_decode(file_get_contents(__DIR__ . '/data/config.json'), true);
+
+        for ($i = 0; $i < $count; $i++) {
+            $docs[] = ['_id' => ['r' => 'tenantLists:batch' . $i, 'c' => 'tenantContexts:DefaultGraph']];
+        }
+
+        $fakeCursor = new ArrayIterator($docs);
+        /** @var \PHPUnit_Framework_MockObject_MockObject|TripodTestConfig $config */
```
rgubby: `$config` -> `$configInstance`

test/unit/mongo/MongoTripodViewsTest.php
```diff
@@ -2276,4 +2276,135 @@ public function testDeleteViewsByViewIdWithTimestamp()
...
+        $docs = [];
+
+        $configOptions = json_decode(file_get_contents(__DIR__ . '/data/config.json'), true);
+
+        for ($i = 0; $i < $count; $i++) {
+            $docs[] = ['_id' => ['r' => 'tenantLists:batch' . $i, 'c' => 'tenantContexts:DefaultGraph']];
+        }
+
+        $fakeCursor = new ArrayIterator($docs);
+        /** @var \PHPUnit_Framework_MockObject_MockObject|TripodTestConfig $config */
```
rgubby: `$config` -> `$configInstance`

src/mongo/delegates/Tables.class.php
rsinger: Probably, but all of this is kind of hairy, anyway.  I think in general there is a lot of consolidation that could be done between the composites to simplify them.

PR #129, opened 2018-08-31
Provide option to return a cursor for table row results

rsinger:
Since table rows are probably the most likely 'bulk' results operation, this branch gives the option to replace the `results` value with a `\MongoDB\Driver\Cursor` object, rather than an array.  The returned documents are identical to the previous behavior.

Also added is an option to suppress getting counts: with the cursor, it's not as necessary to do paging, so doing a count is an unnecessary performance hit.


test/unit/mongo/MongoTripodTablesTest.php
```diff
@@ -1839,4 +1839,70 @@ public function testDeleteTableRowsByTableIdWithTimestamp()
...
+                    'c' => 'tenantContexts:DefaultGraph'
+                ],
+                '_impactIndex' => [
+                    [ 'r' => 'http://talis.com/modules/xmen-004', 'c' => 'tenantContexts:DefaultGraph' ],
+                    [ 'r' => 'http://talis.com/schools/xmen-001', 'c' => 'tenantContexts:DefaultGraph' ]
+                ],
+                'code' => 'XMEN-004',
+                'nodeUrl' => 'http://talis.com/modules/xmen-004',
+                'name' => 'Psychology: Living with The Voices',
+                'description' => 'Professor Deadpool will attempt to give you the ability to embrace your mental disorder and use it to your advantage. Whether you suffer from Schizophrenia, Dissociative Identity Disorder or plain old Comic Awareness, Wade Wilson has probably suffered through it himself. And while Deadpool can\'t solve your problem, he can teach you how to make it one of your most marketable qualities.',
```
rgubby: 👏

PR #132, opened 2019-09-17
Remove ref to getStoreStatPath

rgubby:
If you have a statsd host configured, whenever there is a request to send stats, we work out which stats paths to send data to.

As well as stats about the current operation, in a pretty straight forward `myapp.tripod.MONGO_SOMETHING` we also send `myapp.tripod.group_by_db.db_name.MONGO_SOMETHING`.

There are known issues with statsd sending this data in a blocking fashion, rather than non-blocking, so having more data, that is verbose and as it turns out, really not that useful (we have been blacklisting these stats for years), makes the app run slower.

By reducing the amount of data, we reduce the [number of loops](https://github.com/talis/tripod-php/blob/master/src/classes/StatsD.class.php#L109) we do, which means that we block the operation for less. This means a more performant app.

### DIFF
--- a/src/classes/StatsD.class.php
+++ b/src/classes/StatsD.class.php
@@ -251,7 +251,7 @@ class StatsD implements ITripodStat
      */
     protected function getStatsPaths()
     {
-        return(array_values(array_filter(array($this->getAggregateStatPath()))));
+        return array_values(array_filter([$this->getAggregateStatPath()]));
     }
 
     /**
--- a/src/classes/StatsD.class.php
+++ b/src/classes/StatsD.class.php
@@ -251,24 +251,7 @@ class StatsD implements ITripodStat
      */
     protected function getStatsPaths()
     {
-        return(array_values(array_filter(array($this->getStoreStatPath(), $this->getAggregateStatPath()))));
-    }
-
-    /**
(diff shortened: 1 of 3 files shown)


kiyanwang:
This will definitely stop sending the additional stats. I know we've discussed whether this should be optional or driven through some configuration but it seems there isn't really an appetite for that so I'm happy with this for now.

src/classes/StatsD.class.php
```diff
@@ -251,24 +251,7 @@ public function setPivotValue($pivotValue)
...
-        if(isset($this->pivotValue))
-        {
-            if(!empty($this->prefix))
-            {
-                $path = $this->prefix . '.';
-            }
-            $path .= STAT_CLASS . '.' . STAT_PIVOT_FIELD . '.' . $this->pivotValue;
-        }
-        return $path;
+        return(array_values(array_filter(array($this->getAggregateStatPath()))));
```
kiyanwang: space btw `return` and `(`

rsinger: Tripod requires `>= 5.5` so this can just be:
`return array_values(array_filter([$this->getAggregateStatPath()]));`

PR #133, opened 2020-03-17
Added 7.1 to travis

rgubby:
Just for testing, nothing to see here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 sudo: required
 php:
+- 7.1
 - 7.0
 - 5.6
 - 5.5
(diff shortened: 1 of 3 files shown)


malcyL:
The travis builds of talis-php started failing with this error recently. 

See https://github.com/talis/talis-php/pull/16

> This PR installs ant to fix the 5.6 to 7.2 inclusive builds.
> The build image is pinned to trusty which is needed to be able to install PHP 5.5. (This may mean that ant is installed by default without specifying it explicitly, but we are using it, and when we do move to Xenial or later it might not be there. We're using it - lets be explicit. )

PR #134, opened 2020-11-11
PLAT-4051 pin composer version tripod-php

Aardeedev:
talis/platform#4051

pin composer version to 1.10.17

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -50,7 +50,7 @@
         <echo>Installing composer...</echo>
         <exec executable="sh">
             <arg value="-c"/>
-            <arg value="curl -sS https://getcomposer.org/installer | php"/>
+            <arg value="curl -sS https://getcomposer.org/installer | php -- --version 1.10.17"/>
         </exec>
     </target>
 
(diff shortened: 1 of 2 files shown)


build.xml
```diff
@@ -21,12 +21,6 @@
         <delete dir="${output.dir}" />
         <mkdir dir="${output.dir}" />
 
-        <!-- Check for updates to Composer -->
-        <exec executable="php" failonerror="true">
-            <arg value="composer.phar"/>
-            <arg value="self-update"/>
-        </exec>
```
Aardeedev, line 28: We no longer need this as we are pinning to the lates v1 release (1.10.17)

rgubby (commented):
You might want to pick these 2 changes:

https://github.com/talis/tripod-php/pull/133/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

and 

https://github.com/talis/tripod-php/pull/133/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

They look build related and might help!

rgubby:
@Aardeedev You will need to do a rebase on this branch to be able to start getting green builds.

build.xml
```diff
@@ -21,23 +21,17 @@
...
             <arg value="install"/>
         </exec>
 
         <!-- Update dependencies through Composer -->
-        <exec executable="php" failonerror="true">
+        <!-- <exec executable="php" failonerror="true">
             <arg value="composer.phar"/>
             <arg value="update"/>
-        </exec>
+        </exec> -->
```
mrliptontea, line 34: Given that we don't include `composer.lock` (it's a library so it's correct) we don't need to update so I'm happy with removing this.

mrliptontea (approved):

PR #139, opened 2021-12-17
90%: Migrate build to circleci

rsinger:
Our build _had_ been in Travis-CI, but that hasn't worked since they deprecated travis-ci.org.

This moves it over to Circle and sets up a build matrix from PHP 5.5 - 7.2


docker/Dockerfile-php72
```diff
@@ -0,0 +1,10 @@
+FROM php:7.2-cli
+
+ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
+
+RUN chmod +x /usr/local/bin/install-php-extensions && \
+    install-php-extensions pcntl redis mongodb-1.4.4 && \
```
rsinger, line 6: This one requires a newer mongodb extension. 1.1.9 doesn't work with 7.2 and 1.4.4 is the last version that works with MongoDB 2.6

mikerudd (approved):
All looks good to me. Nice use of the circle matrix feature, not seen that before.

PR #140, opened 2022-09-05
PLAT-5896 Add support for setting Resque Redis database

mrliptontea:
* https://github.com/talis/platform/issues/5896
* We need to be able to specify a custom Redis database number for Resque.


test/unit/mongo/MongoTripodConfigTest.php
```diff
@@ -1707,6 +1707,17 @@ public function testGetResqueServer()
         $this->assertEquals(getenv(MONGO_TRIPOD_RESQUE_SERVER), \Tripod\Mongo\Config::getResqueServer());
     }
 
+    public function testGetResqueDatabase()
+    {
+        \Tripod\Mongo\Config::setValidationLevel(\Tripod\Mongo\Config::VALIDATE_MAX);
```
mrliptontea, line 1712: I don't think this is needed? Copied from the test above but that also looks outdated.

src/mongo/Config.class.php
```diff
@@ -2110,6 +2110,14 @@ public static function getResqueServer()
         return $resqueServer;
     }
 
+    /**
+     * @return int|string
+     */
+    public static function getResqueDatabase()
+    {
+        return self::getenv(RESQUE_DATABASE, 0);
```
mrliptontea, line 2118: I don't think we should warn about it missing because it is optional and has a default.

mrliptontea:
Turns out Resque PHP supports DSN as a server parameter
```
redis[s]://[pass@][ip|host|socket[:port]][/db-index]
```

> A database number to select. However, if we find a valid database number in the DSN the
DSN-supplied value will be used instead and this parameter is ignored.

We're going to do that instead.

PR #143, opened 2023-10-24
PLT-72 MongoDB extension and library upgrade

mrliptontea:
* [PLT-72](https://techfromsage.atlassian.net/browse/PLT-72)
* Upgrade MongoDB extension and library to match versions that work with PHP 5.5 - 7.3.
* Remove builds against PHP 5.6, 7.1, 7.2 as we don't use them.
* Refactor CircleCI config.
* Remove unused build files.


stephenrob (approved):

kiyanwang (approved):
That looks good.

mrliptontea:
Just so I can leave that somewhere - I wanted to also include PHP 7.4 but with current code it won't work.

I haven't looked further than that, but an issue I found while running unit tests is that `Tripod\Mongo\Jobs\JobBase` class, and classes inheriting from it, on numerous occasions are referencing `$this->job` and `$this->args` properties but they're never set anywhere and are not defined in the class.

The only way it kind of works in PHP 7.3 still is there seems to be some nuance around setting such properties as if they're arrays vs if they're just a string value. Better explained with code below:

```sh
# Works ('foo' is created as stdClass with 'bar' array)
php73 -r '$obj = new \stdClass(); $obj->foo->bar["baz"] = uniqid(); var_dump($obj);'
# Warning: Creating default object from empty value
php74 -r '$obj = new \stdClass(); $obj->foo->bar["baz"] = uniqid(); var_dump($obj);'

# BUT!

# Warning: Creating default object from empty value
php73 -r '$obj = new \stdClass(); $obj->foo->bar = uniqid(); var_dump($obj);'
# Warning: Creating default object from empty value
php74 -r '$obj = new \stdClass(); $obj->foo->bar = uniqid(); var_dump($obj);'
```

Granted, this produces _only_ a warning, but whether that stops the app depends on how errors are handled within the app. PHPUnit fails on this, for example.

composer.json
```diff
@@ -17,14 +17,18 @@
...
         "php" : ">=5.5",
-        "semsol/arc2": "v2.2.4",
-        "chrisboulton/php-resque": "dev-master#98fde571db008a8b48e73022599d1d1c07d4a7b5",
-        "monolog/monolog" : "~1.13",
-        "mongodb/mongodb": "1.0.5"
+        "semsol/arc2": "2.2.6",
+        "monolog/monolog" : "~1.13"
     },
     "require-dev": {
+        "resque/php-resque": "v1.3.4",
```
rsinger, line 30: I would prefer to keep this in alignment with RL: https://github.com/talis/rl-app/blob/master/composer.json#L67, namely because of the way it introduces failed-queues

rsinger (approved):
I'm ok with everything here but a little itchy about the resque change

composer.json
mrliptontea: OK so buckle up for this :roller_coaster:  

The `chrisboulton/php-resque` referenced here in tripod, is the OG PHP resque implementation, which was later moved to the official resque repository, as `resque/php-resque` package. It literally is the same thing, but still maintained (also, the maintained version includes some fixes that we should be interested in, except we couldn't update to them due to PHP version constraint).

However, what RL references in its composer.json, isn't the same `chrisboulton/php-resque` - it's [our fork](https://github.com/talis/php-resque) where we indeed implemented `RedisMultipleQueues` but, because we didn't change the name of the package, it looks like the OG one, except composer prioritises custom repositories, so it pulls our fork in RL.

So we're already in a situation where it's a different php-resque between tripod and RL. And this is precisely my point - tripod is a library, and it shouldn't impose the configuration of another library on the application (unless it's absolutely needed).

In DC we use the official package with [our package](https://github.com/talis/resque-multiple-failure-queues) that provides the custom job failure backend:

https://github.com/talis/dc-app/blob/2baaedac/composer.json#L67-L68
https://github.com/talis/dc-app/blob/2baaedac/conf/queues.inc.php#L56

PR #144, opened 2023-12-01
PLT-71 Fix PHP type issues

mrliptontea:
* [PLT-71](https://techfromsage.atlassian.net/browse/PLT-71)
* We're seeing some PHP errors around Tripod in production, some of them become a bigger problem in later version of PHP. Also, in the newer MongoDB driver the BSON classes become iterable, which changes their behaviour within a `foreach`, which also leads to PHP errors when it's not accounted for.

### DIFF
--- a/src/classes/ExtendedGraph.class.php
+++ b/src/classes/ExtendedGraph.class.php
@@ -54,7 +54,7 @@ class ExtendedGraph
     var $_labeller;
 
     /**
-     * @param string|\Tripod\ExtendedGraph $graph
+     * @param string|array $graph
      */
     public function __construct($graph=null){
         $this->_labeller = new Labeller();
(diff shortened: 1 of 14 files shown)


src/classes/ExtendedGraph.class.php
```diff
@@ -549,7 +549,10 @@ public function from_rdfxml($rdfxml, $base='') {
     public function from_json($json) {
         if ($json) {
             $this->remove_all_triples();
-            $this->_index = json_decode($json, true);
+            $index = json_decode($json, true);
+            if (is_array($index)) {
```
kiyanwang, line 553: I was going to ask why we need to check if its an array, then I remembered that `$this->_index` is initalised as an empty array, and json_decode here will return either an associative array or `null`. I'm guessing as it was it was generating warnings.

PR #146, opened 2024-01-17
PLT-244 Fix Tripod Resque event listeners

mrliptontea:
* Related to [PLT-244](https://techfromsage.atlassian.net/browse/PLT-244)

An error was seen in production:

```
PHP Fatal error:  Uncaught Error: Call to undefined method RisImportJob::errorLog() in /var/rl-app/vendor/talis/tripod-php/src/mongo/base/JobBase.class.php:107
```

This was caused by the fact that Tripod's JobBase class' constructor registers Resque `onFailure` event listener, so it's possible that this function will get called by a non-tripod worker in case of an error.

Two possible fixes:

1. (Currently implemented, non-breaking) make sure event listeners ignore calls for jobs that don't inherit JobBase.
2. (Breaking) Remove JobBase constructor, and move the responsibility for registering listeners to library's consumer.

[PLT-244]: https://techfromsage.atlassian.net/browse/PLT-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mongo/base/JobBase.class.php
+++ b/src/mongo/base/JobBase.class.php
@@ -31,8 +31,8 @@ abstract class JobBase extends \Tripod\Mongo\DriverBase
 
     public function __construct()
     {
-        \Resque_Event::listen('beforePerform', [$this, 'beforePerform']);
-        \Resque_Event::listen('onFailure', [$this, 'onFailure']);
+        \Resque_Event::listen('beforePerform', [self::class, 'beforePerform']);
+        \Resque_Event::listen('onFailure', [self::class, 'onFailure']);
     }
 
     /**
@@ -63,7 +63,12 @@ abstract class JobBase extends \Tripod\Mongo\DriverBase
      */
     public static function beforePerform(\Resque_Job $job)
     {
-        $job->getInstance()->validateArgs();
+        $instance = $job->getInstance();
+        if (!$instance instanceof self) {
+            return;
+        }
+
+        $instance->validateArgs();
     }
 
     public function setUp()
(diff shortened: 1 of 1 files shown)


src/mongo/base/JobBase.class.php
```diff
@@ -31,8 +31,8 @@ abstract class JobBase extends \Tripod\Mongo\DriverBase
 
     public function __construct()
     {
-        \Resque_Event::listen('beforePerform', [$this, 'beforePerform']);
-        \Resque_Event::listen('onFailure', [$this, 'onFailure']);
+        \Resque_Event::listen('beforePerform', [self::class, 'beforePerform']);
+        \Resque_Event::listen('onFailure', [self::class, 'onFailure']);
```
mrliptontea, line 35: Methods are static, so should be called statically.

src/mongo/base/JobBase.class.php
```diff
@@ -96,14 +101,17 @@ public function tearDown()
...
      * Resque event when a job failures
      *
-     * @param \Exception  $e   Exception
-     * @param \Resque_Job $job The failed job
+     * @param \Throwable|\Exception $e   Exception or Error
+     * @param \Resque_Job           $job The failed job
      * @return void
      */
-    public static function onFailure(\Exception $e, \Resque_Job $job)
+    public static function onFailure($e, \Resque_Job $job)
```
mrliptontea, line 108: In PHP 7+ this might be called with an `Error` too, which extends `Throwable`. As the library currently supports PHP 5.5 I can't use that keyword, hence removed the typehint.

kiyanwang (approved):
LGTM

PR #147, opened 2024-01-19
PLT-299 Remove Resque event listeners from JobBase constructor

mrliptontea:
* Related to [PLT-299](https://techfromsage.atlassian.net/browse/PLT-299); this is precursory work to updating php-resque in TARL.
* JobBase class' constructor registers Resque_Event listeners but it's more practical that the consumer application does this. However, as it's required that at least `beforePerform` listener is registerd, it will be a breaking change. On the other hand, RL already does register these listeners, which means they're fired twice on event.
* Declare JobBase properties that are injected by Resque.
* Consolidate PHPUnit configuration files.
* Add test bootstrap file.

[PLT-299]: https://techfromsage.atlassian.net/browse/PLT-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mongo/base/JobBase.class.php
```diff
@@ -15,6 +15,24 @@
...
+     * @var string
+     */
+    public $queue;
+
+    /**
+     * Resque Job
+     * @var \Resque_Job
+     */
+    public $job;
+
```
mrliptontea, line 35: These are injected by Resque_Job

https://github.com/talis/php-resque/blob/1.2/lib/Resque/Job.php#L158-L161

https://github.com/resque/php-resque/blob/v1.3.6/lib/Resque/Job/Factory.php#L27-L29
https://github.com/resque/php-resque/blob/v1.3.6/lib/Resque/Job.php#L178-L179

kiyanwang (approved):
I think that looks ok. 
I wonder the double event firing was one of the reasons why it was so difficult to keep track of these in the Instana (it wont be the root cause but probably did muddy the water)

nice work

PR #150, opened 2024-09-05
PLT-856 Update PHPUnit

mrliptontea:
* PLT-856
* Update PHPUnit
* Drop PHP 5 support
* [PSR-12](https://www.php-fig.org/psr/psr-12/) formatting in tests


kiyanwang (approved):
That looks good. 
I know I often bleet on about separating formatting/styling from functional changes but here I think it was a case of JFDI.  
Nice work.

PR #151, opened 2024-11-01
Adding support for joining on _id in a view

pogotc:
**Background**

Tables allow you to join on an `_id` field, for example: https://github.com/techfromsage/rl-app/blob/ca940e8c1ce84a95079535106b2652fc4ea6b612/environmentconfig/tripod/tenant-tablespecs.json#L1571

But it's not supported in views, even though it would be quite useful.

### DIFF
--- a/src/mongo/delegates/Views.class.php
+++ b/src/mongo/delegates/Views.class.php
@@ -571,9 +571,9 @@ class Views extends CompositeBase
                 {
                     // single value for join
                     $joinUris[] = array(_ID_RESOURCE=>$source[$predicate][VALUE_URI],_ID_CONTEXT=>$contextAlias);
-                }
-                else
-                {
+                } else if($predicate == '_id') {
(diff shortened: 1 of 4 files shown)


src/mongo/delegates/Views.class.php
```diff
@@ -571,9 +571,9 @@ protected function doJoins($source, $joins, &$dest, $from, $contextAlias, $build
                 {
                     // single value for join
                     $joinUris[] = array(_ID_RESOURCE=>$source[$predicate][VALUE_URI],_ID_CONTEXT=>$contextAlias);
-                }
-                else
-                {
+                } else if($predicate == '_id') {
+                    $joinUris[] = array(_ID_RESOURCE => $source[$predicate][_ID_RESOURCE], _ID_CONTEXT => $contextAlias);
```
rsinger, line 575: Since Tripod now requires PHP 7.3+, we can ditch `array(...)`, but I appreciate that consistency may be more important

rsinger (approved):
Glad I decided not to take that bullet for you!

PR #152, opened 2024-11-28
TA-1913 - Support Mongo index options in Tripod config

astilla:
### Fix
* Cater for a new (non-breaking) index format where there are two arrays instead of one - the second array allows passing of standard Mongo index options (e.g. sparse=>true, unique=>true)
* Add new tests to check 
* Tested the dependency manually inside local TARL to prove the index gets generated correctly and no other fallout.
* See associated TARL PR where we use the new index definitions to create a sparse/unique index on `personaUsers.guid`: https://github.com/techfromsage/rl-app/pull/4461

### Approach taken
* We drove this off the existing tests and copious amounts of temporary logging.  Adding more tests that support that new two array format (where second array is standard Mongo index options)
* Once all passing in tests, copied this code into TARL's vendor directory and updated the `tenant-pods.json` file in TARL to add a new index onto CBD_users `personaUsers.guid` that also had sparse=>true & unique=>true.
* Reloaded TARL data, no indexes present (this also generates the tripod config)
* Kicked off Setup Console "ensureIndexes"
* Looked in TARL and Tripod logs for errors (non present) and also checked CBD_users indexes - new index was present with the correct index options.
* As a safety test we also tried the index _without_ the `sparse=>true` and the index generation failed as there are many documents in local TARL that have no persona guid set.  

### Issues
It's worth noting a few issues encountered along the way with this...
* TARL's `docker-compose.yml` has a `:delegated` suffix on the `vendor` volume mount, effectively caching changes so when you copy new code in it wasn't being picked up.  We temporarily removed this to prove our changes, but _sometimes_ it still didn't pick up changes quickly enough.  The only reliable option was to stop/start the RL containers and then code in `vendor` was correctly reflected.
* It would have been nice to have left a logger statement in at the point of creating the index, just so we can see the options, but the logger isn't available in the IndexUtils class.

### Example 
From a `db.CBD_users.getIndexes()` in the local Mongo `life` database, the new index is present:
Screenshot 2024-12-03 at 10 21 56

### DIFF
--- a/src/mongo/Config.class.php
+++ b/src/mongo/Config.class.php
@@ -254,29 +254,31 @@ class Config implements IConfigInstance
                             $todoKeys = array_keys($indexFields);
                             if (is_numeric($todoKeys[0])) {
                                 error_log('new format config - two arrays' . json_encode($todoKeys[0]));
+                                $todoIndexFields = $indexFields[0];
                             } else {
                                 error_log('old format config' . json_encode($todoKeys[0]));
+                                $todoIndexFields = $indexFields;
                             }
 
                             // If we have the new config, it doesn't like this cardinality check
                             // so bypass it for now and come back to it later once discussed
 
                             // check no more than 1 indexField is an array to ensure Mongo will be able to create compound indexes
-                            // if (count($indexFields) > 1) {
-                            //     $fieldsThatAreArrays = 0;
-                            //     foreach ($indexFields as $field => $fieldVal) {
-                            //         $cardinalityField = str_replace('.value', '', $field);
-                            //         if (!array_key_exists($cardinalityField, $this->cardinality[$storeName][$podName]) ||
-                            //             $this->cardinality[$storeName][$podName][$cardinalityField] != 1) {
-                            //             $fieldsThatAreArrays++;
-                            //         }
-                            //         if ($fieldsThatAreArrays > 1) {
-                            //             throw new \Tripod\Exceptions\ConfigException("Compound index $indexName has more than one field with cardinality > 1 - mongo will not be able to build this index");
-                            //         }
-                            //     }
-                            // }
+                            if (count($todoIndexFields) > 1) {
+                                $fieldsThatAreArrays = 0;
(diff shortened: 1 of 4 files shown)


src/mongo/Config.class.php
```diff
@@ -247,11 +247,26 @@ protected function loadConfig(array $config)
...
                         $this->indexes[$storeName][$podName] = [];
+
+                        error_log('Config indexes: '.json_encode($podConfig["indexes"]));
                         foreach ($podConfig["indexes"] as $indexName => $indexFields) {
+
+                            $this->indexes[$storeName][$podName][$indexName] = $indexFields;
+                            error_log('Index for store: '.$storeName.' - '.json_encode($indexFields));
+
+                            $indexKeys = array_keys($indexFields);
+                            if (is_numeric($indexKeys[0])) {
```
mikerudd, line 258: [question] Do we need to guard against `$indexKeys[0]` being null?

mikerudd, line 258: [question] could this logic be pulled out to a well named function to help explain it's intention?

mikerudd (commented):
This seems ok to me. I'll say it picked my brain for a bit but I think what you have will do the job.

test/unit/mongo/IndexUtilsTest.php
```diff
@@ -407,7 +465,7 @@ protected function getCollectionForCBDShouldNeverBeCalled($mockConfig)
      * @param bool $background create indexes in the background
      * @return void
      */
-    protected function oneCustomAndThreeInternalTripodCBDIndexesShouldBeCreated($mockCollection, $background = true)
+    protected function oneCustomAndThreeInternalTripodCBDIndexesShouldBeCreated($mockCollection, $background = true, array $indexOptions = [])
```
rsinger, line 468: These methods... just crazy

rsinger (commented):
This looks good to me so far

rsinger (approved):
One formatting comment. I still don't really understand the relationship between cardinality and indexes but that's also a problem for another time

PR #156, opened 2026-02-27
Modernise code

mrliptontea:
- Precursor to PLT-951
    - Modernise codebase with PHPStan, PHP CS Fixer, and Rector
    - Drop PHP 7.3
    - Use PHP 7.4 features
    - Strict types where possible
    - Resolve `todo: use ObjectId->getTimestamp()`

<details><summary>Used <code>rector.php</code></summary>

```php
<?php

declare(strict_types=1);

use Rector\CodeQuality\Rector\Empty_\SimplifyEmptyCheckOnEmptyArrayRector;
use Rector\CodeQuality\Rector\Identical\FlipTypeControlToUseExclusiveTypeRector;
use Rector\CodingStyle\Rector\Catch_\CatchExceptionNameMatchingTypeRector;
use Rector\CodingStyle\Rector\ClassMethod\MakeInheritedMethodVisibilitySameAsParentRector;
use Rector\CodingStyle\Rector\PostInc\PostIncDecToPreIncDecRector;
use Rector\Config\RectorConfig;
use Rector\Naming\Rector\Assign\RenameVariableToMatchMethodCallReturnTypeRector;
use Rector\Naming\Rector\Class_\RenamePropertyToMatchTypeRector;
use Rector\Naming\Rector\ClassMethod\RenameParamToMatchTypeRector;
use Rector\Naming\Rector\ClassMethod\RenameVariableToMatchNewTypeRector;
use Rector\Naming\Rector\Foreach_\RenameForeachValueVariableToMatchExprVariableRector;
use Rector\Privatization\Rector\Class_\FinalizeTestCaseClassRector;
use Rector\Strict\Rector\Empty_\DisallowedEmptyRuleFixerRector;
use Rector\TypeDeclaration\Rector\ClassMethod\ParamTypeByMethodCallTypeRector;
use Rector\TypeDeclarationDocblocks\Rector\Class_\AddReturnDocblockDataProviderRector;
use Rector\TypeDeclarationDocblocks\Rector\ClassMethod\AddParamArrayDocblockFromDimFetchAccessRector;
use Rector\TypeDeclarationDocblocks\Rector\ClassMethod\DocblockReturnArrayFromDirectArrayInstanceRector;

return RectorConfig::configure()
    ->withPaths([
        __DIR__ . '/scripts',
        __DIR__ . '/src',
        __DIR__ . '/test',
    ])
    ->withBootstrapFiles([
        __DIR__ . '/src/tripod.inc.php',
    ])
    ->withPhpSets(php74: true)
    ->withPreparedSets(
        deadCode: true,
        codeQuality: true,
        codingStyle: true,
        typeDeclarations: true,
        typeDeclarationDocblocks: true,
        privatization: true,
        naming: true,
        instanceOf: true,
        earlyReturn: true,
        rectorPreset: true,
        phpunitCodeQuality: true,
    )
    ->withSkip([
        AddReturnDocblockDataProviderRector::class,
        CatchExceptionNameMatchingTypeRector::class,
        DisallowedEmptyRuleFixerRector::class,
        FinalizeTestCaseClassRector::class,
        MakeInheritedMethodVisibilitySameAsParentRector::class,
        ParamTypeByMethodCallTypeRector::class,
        PostIncDecToPreIncDecRector::class,
        RenameForeachValueVariableToMatchExprVariableRector::class,
        RenameParamToMatchTypeRector::class,
        RenamePropertyToMatchTypeRector::class,
        RenameVariableToMatchMethodCallReturnTypeRector::class,
        RenameVariableToMatchNewTypeRector::class,
        SimplifyEmptyCheckOnEmptyArrayRector::class,
        FlipTypeControlToUseExclusiveTypeRector::class,
        AddParamArrayDocblockFromDimFetchAccessRector::class,
        DocblockReturnArrayFromDirectArrayInstanceRector::class,
    ])
;
```
</details>


src/classes/ChangeSet.php
```diff
@@ -65,9 +78,7 @@ public function __construct(array $a)
                     $parser->parse(false, $a[$rdf]);
                     $a[$rdf] = $parser->getSimpleIndex(0);
                 } elseif (
-                    is_array($a[$rdf])
-                    and isset($a[$rdf][0])
```
astilla, line 69: I'd long since forgotten `and` existed in PHP...!

src/classes/ExtendedGraph.php
```diff
@@ -1869,17 +1811,15 @@ private function _to_arc_index(&$index)
...
-     * @param string $o
-     * @param string $type
-     *
-     * @return array
+     * @param TriplePredicate $p
+     * @param ObjectValue     $o
+     * @param ObjectType      $type
      */
-    private function get_subjects_where($p, $o, $type)
+    private function get_subjects_where(string $p, $o, string $type): array
```
astilla, line 1818: Why does the docblock type it as a `TriplePredicate` but in the function it types it as a `string` ?

src/classes/ExtendedGraph.php
```diff
@@ -1894,19 +1834,19 @@ private function get_subjects_where($p, $o, $type)
...
     /**
-     * @param string $s
-     * @param string $o
-     * @param string $type
+     * @param TripleSubject        $s
+     * @param ObjectValue          $o
+     * @param 'literal'|'resource' $type
      */
-    private function add_to_sequence($s, $o, $type = 'resource')
+    private function add_to_sequence(string $s, $o, string $type = 'resource'): void
```
astilla, line 1841: Same question as above, but `TripleSubject` vs. `string`

phpstan.neon
```diff
@@ -0,0 +1,17 @@
+parameters:
+  level: 5
```
astilla: Woop woop

astilla (commented):
Well...it's massive... 🤣 
Looks like some good automated updates.
I've left a couple of questions I wasn't sure about a difference in typing between docblock and function param definition.
Other than than it looks sane but being so big I've somewhat skimmed it.

src/classes/ExtendedGraph.php
mrliptontea: Types like `TriplePredicate`, `ObjectValue` are [PHPStan type aliases](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases), defined here:

https://github.com/talis/tripod-php/blob/2adc87526ae2a47d9fa538bd430c693ac7ad4e6f/src/classes/ExtendedGraph.php#L12-L17

(though language servers in PHPStorm, Intelephense in VSCode, etc. also interpret them and provide completions based on them)

I'm using `TriplePredicate` for consistency, even though it translates to just string. This helps make a better sense of what's inside the object, which is especially visible here:

https://github.com/talis/tripod-php/blob/2adc87526ae2a47d9fa538bd430c693ac7ad4e6f/src/classes/ExtendedGraph.php#L19

I think this enhances the reading experience as you can quickly find where predicates are referenced, such as here:

https://github.com/talis/tripod-php/blob/2adc87526ae2a47d9fa538bd430c693ac7ad4e6f/src/classes/ExtendedGraph.php#L954-L956

So you know it's not just an array being returned, it's an array of predicates.

src/classes/ExtendedGraph.php
mrliptontea: See above

scripts/mongo/createSearchDocuments.php
```diff
@@ -42,32 +42,27 @@ function showUsage($scriptName)
     echo $help;
 }
 
-if (empty($options) || isset($options['h']) || isset($options['help'])
+if (
+    $options === [] || $options === false || isset($options['h']) || isset($options['help'])
```
mike-dean-talis, line 49: comment: I am surprised that `$options === []` works as intended to be honest. `empty` does check for false-y values, so I am surprised this is necessary.

scripts/mongo/createSearchDocuments.php
```diff
@@ -85,23 +80,15 @@ function generateSearchDocuments($id, $specId, $storeName, $stat = null, $queue
...
 } else {
     $specId = null;
 }
 
-if (isset($options['i']) || isset($options['id'])) {
-    $id = $options['i'] ?? $options['id'];
-} else {
-    $id = null;
-}
+$id = isset($options['i']) || isset($options['id']) ? $options['i'] ?? $options['id'] : null;
```
mike-dean-talis: comment: I wonder if we can make it simpler than this:

```php
$id = $options['i'] ?? $options['id'] ?? null; 
```

We seem to do similar elsewhere, so I won't go through them all (I'll be here forever) but you can get my point.

scripts/mongo/createTables.php
```diff
@@ -44,32 +44,31 @@ function showUsage()
...
  */
-function generateTables($id, $tableId, $storeName, $stat = null, $queue = null)
+function generateTables(?string $id, string $tableId, string $storeName, ?ITripodStat $stat = null, ?string $queue = null): void
 {
     $tableSpec = Config::getInstance()->getTableSpecification($storeName, $tableId);
     if (array_key_exists('from', $tableSpec)) {
         Config::getInstance()->setMongoCursorTimeout(-1);
 
-        echo "Generating {$tableId}";
+        echo 'Generating ' . $tableId;
```
mike-dean-talis, line 72: question: I am not entirely sure what was "bad" about the previous syntax? We do a lot of interleaved strings like this in JS, I just wonder what makes it "modern" to go backwards? Anyway... enough of my opinionated waffle.

scripts/mongo/detectNamespaces.php
```diff
@@ -107,66 +113,9 @@
...
-        $prevChar = $char;
-    }
-
-    return $result;
-}
-
-$json = json_encode(['namespaces' => $config['namespaces']]);
-
-echo indent($json);
+echo json_encode(['namespaces' => $config['namespaces']], JSON_PRETTY_PRINT);
```
mike-dean-talis, line 121: praise: This one made me laugh out loud. Why have 50 lines when you can have one? 👍

scripts/mongo/loadTriples.php
```diff
@@ -46,21 +44,22 @@ function load(TriplesUtil $loader, $subject, array $triples, array &$errors, $po
...
+    if ($i % 250000 === 0) {
         echo 'Memory: ' . memory_get_usage() . "\n";
     }
 
     $line = rtrim($line);
     $parts = preg_split('/\s/', $line);
     $subject = trim($parts[0], '><');
 
-    if (empty($currentSubject)) { // set for first iteration
+    if ($currentSubject === '' || $currentSubject === '0') { // set for first iteration
```
mike-dean-talis: comment: This PR makes me wonder if there is something inherently wrong with `empty`? As it seems to be avoided in favour of direct comparisons that I wouldn't usually use.

src/classes/Labeller.php
```diff
@@ -320,39 +324,45 @@ public function qname_to_uri($qname)
...
+                preg_match('~^[a-zA-Z][a-zA-Z0-9\-]+$~', $parts[$i])
+                && !isset($this->_ns[$parts[$i]])
+                && $parts[$i] != 'schema'
+                && $parts[$i] != 'ontology'
+                && $parts[$i] != 'vocab'
+                && $parts[$i] != 'terms'
+                && $parts[$i] != 'ns'
+                && $parts[$i] != 'core'
+                && strlen($parts[$i]) > 3
+            ) {
```
mike-dean-talis, line 365: comment: This is my favourite `if` statement. I have no idea what it's really doing or why.

No change necessary just... man. I despair.

src/mongo/serializers/NQuadSerializer.php
```diff
@@ -64,12 +71,9 @@ public function getTerm($v)
...
 
     /**
-     * @param array  $index
-     * @param string $context
-     *
-     * @return string
+     * @param array $index
      */
-    public function getSerializedIndex($index, $context)
+    public function getSerializedIndex($index, ?string $context): string
```
mike-dean-talis: question: Can we not type `$index` as an `array`? Maybe I have missed something obvious.

src/mongo/Driver.php
```diff
@@ -556,26 +493,18 @@ public function getETag($resource, $context = null)
...
-            $eTag = str_pad(number_format($seconds - floor($seconds), 6), 10, '0', STR_PAD_RIGHT) . ' ' . floor($seconds);
+            return '';
         }
 
-        return $eTag;
+        $milliseconds = (int) $lastUpdatedDate->__toString();
+        $seconds = intdiv($milliseconds, 1000);
+        $fraction = ($milliseconds % 1000) / 1000;
+
+        return sprintf('%.8f %d', $fraction, $seconds);
```
mike-dean-talis, line 504: praise: I _think_ this is doing the same work. I'm hoping tests would catch it if not. This makes the process more readable at least.

test/unit/mongo/ConfigGeneratorTest.php
```diff
@@ -24,33 +26,34 @@ protected function setUp(): void
...
-    public function testCreateFromConfig()
+    public function testCreateFromConfig(): void
     {
         /** @var TestConfigGenerator $instance */
         $instance = Config::getInstance();
         $this->assertInstanceOf(TestConfigGenerator::class, $instance);
         $this->assertInstanceOf(Tripod\Mongo\Config::class, $instance);
         $this->assertInstanceOf(ITripodConfigSerializer::class, $instance);
-        $this->assertEquals(
+        $this->assertSame(
```
mike-dean-talis, line 36: praise: Bless your soul. `assertSame`.

test/unit/mongo/EnsureIndexesTest.php
```diff
@@ -31,47 +33,41 @@ protected function setUp(): void
...
-        return [
-            ['tripodConfig', 'tripodConfig or tripodConfigGenerator'],
-            ['storeName'],
-            ['reindex'],
-            ['background'],
-        ];
+        yield ['tripodConfig', 'tripodConfig or tripodConfigGenerator'];
+        yield ['storeName'];
+        yield ['reindex'];
+        yield ['background'];
```
mike-dean-talis, line 59: praise: Another thing I didn't know PHP had... iterators. Well, I hope to add that one to my repertoire.

mike-dean-talis (commented):
I reached the end. Holy moly. That's 2 hours I won't get back.

This generally looks good to me although I have a few questions and observations along the way.

scripts/mongo/createSearchDocuments.php
mrliptontea: This is on Rector: https://getrector.com/rule-detail/disallowed-empty-rule-fixer-rector.

Since PHP's `getopt` can return `array|false` that's what's it checking for.

From type safety standpoint this make more sense, since the definition of what "empty" means in PHP is quite loose...

```console
$ php -r 'var_dump(empty("0"));'
bool(true)
```

scripts/mongo/createSearchDocuments.php
mrliptontea: True, I missed this. Automagic tools can only take us so far I guess...

scripts/mongo/createTables.php
mrliptontea: Again, Rector at play: https://getrector.com/rule-detail/encapsed-strings-to-sprintf-rector

I believe it's about consistency:

- In PHP, unlike JS, `{}` is limited, and only works with variables:
    ```php
    echo "{$var}";
    echo "{$obj->prop}";
    echo "{$obj->method()}";
    echo "{$obj::$staticProp}";
    echo "{$obj::static()}";
    # But you can't e.g.:
    echo "{func()}";
    echo "{CONSTANT}";
    echo "{SomeClass::static()}";
    ```

- There is also cursed, [deprecated](https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated), `"${var}"` syntax.

scripts/mongo/loadTriples.php
mrliptontea: See my comment above; but yes, Rector seem to prefer to avoid `empty` since it's a catch-all for `null`, `false`, `''`, `'0'`, `0`, `0.0`, `[]`. Converting to an explicit comparison forces you to think about types and possible values.

That said: 119a88fc7e43f9ef8fc9ca3a6fee89acd86371ac

scripts/mongo/createSearchDocuments.php
mrliptontea: See bfc8602eabd1affc675c3ce31780a984ea5898b9

src/mongo/Driver.php
mrliptontea: https://github.com/talis/tripod-php/blob/477f576156ab2617c616f3bff110823833d958ae/test/unit/mongo/MongoTripodDriverTest.php#L2278

Though why `microtime` format was chosen in the first place evades me...

test/unit/mongo/EnsureIndexesTest.php
mrliptontea: Generators are pretty neat.

But for writing tests with many cases wait for PHP attributes!
https://github.com/talis/depot/blob/b2d73c12/server/tests/unit/VaultTest.php#L19-L25

src/classes/Labeller.php
mrliptontea: We'll never know. It dates back to the `Initial commit`
https://github.com/talis/tripod-php/blame/428acd7e190e2427a8ee8dce724afc79b0ea249f/src/classes/Labeller.class.php#L345

src/mongo/serializers/NQuadSerializer.php
mrliptontea: It's not you, it'm me (and Rector) who missed this.

Yes, we can type it: 4a8479e4c18704985dcff7fc0f00c372c0a6b816

mikerudd (approved):

PR #158, opened 2026-03-19
PLT-951 Add PHP 8 support

mrliptontea:
- PLT-951
- Add PHP 8 support
- Require ext-mongodb >= 1.7.0
  - Enforce that `$readPreference` is a string
  - Switch from `ReadPreference->getMode()` to `getModeString()`
- Require `psr/log` instead of Monolog
  - Monolog is now a dev-only package
- Add Resque compatibility layer
  - Allow library to work against current latest release but also development which includes namespace changes and also requires that job classes extend `Resque\Job\Job`.
- Decouple `Tripod\Mongo\Jobs\JobBase` from `Tripod\Mongo\DriverBase`
  - Extract `LoggerTrait` from `DriverBase`
  - Few methods from `DriverBase` used by jobs were re-implemented.
  - Make `Tripod\Mongo\Jobs\JobBase` extend `Resque\Job\Job`.
- Bump arc2 library
  - Allow versions 2 & 3.


mike-dean-talis (approved):
Looks good to an untrained tripod-ist.

